PR #2, opened 2017-08-31
Add the k8s_object rule.

mattmoor:
This contains just enough to define a k8s_object rule that effectively passes through the input object.

Subsequent changes will expand the surface of this to support more meaningful operations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,13 @@ Travis CI | Bazel CI
 
 ## Rules
 
-* Coming Soon!
+* [k8s_object](#k8s_object)
 
 ## Overview
 
-This is a placeholder repository for holding rules to interact with Kubernetes
-constructs.
+
+This repository contains rules for interacting with Kubernetes
+configurations / clusters.
 
 ## Setup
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -30,6 +31,80 @@ load("@io_bazel_rules_k8s//k8s:k8s.bzl", "k8s_repositories")
...
+## Examples
+
+### Basic "deployment" objects
+
+```python
+load("@io_bazel_rules_k8s//k8s:object.bzl", "k8s_object")
+
+k8s_object(
+  name = "dev",
+  kind = "deployment",
```
dlorenc: Do you need a `version` field in k8s object at all? The docs still say to use apps/v1beta1 for Deployment: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

dlorenc: Ah I see, the user is expected to set that themselves in the object. What is the "kind" used for then? It doesn't look like anything right now.

mattmoor: Nothing yet, it is ultimately used for `kubectl` subcommands.

dlorenc: Ah, it just seems a little weird to have it here and in the file itself. Is there anything coming to automatically extract it from the file? What if it doesn't match?

mattmoor: IDK if we'll extract it, but we may add validation that the object we're dealing with matches what's in the file (if such a thing is possible).  Ultimately, with CRDs, there doesn't need to be any correlation between the `kind` in the resource and the command-line keyword.

PR #5, opened 2017-09-02
Add bazel run tag resolution

mattmoor:


mattmoor:
This builds on https://github.com/bazelbuild/rules_k8s/pull/3

mattmoor:
This is now rebased.

.travis.yml
```diff
@@ -33,9 +33,10 @@ install:
   - go get -u github.com/bazelbuild/buildifier/buildifier
 
 script:
-  # Check that all of our samples build
+  # Check that all of our tools and samples build
   - bazel clean && bazel build //...
-  # We have no Bazel-based testing currently.
+  # Check that all of our tests pass
+  - bazel clean && bazel test //...
```
dlorenc: I noticed we're doing this in a few other repos, but is there any reason to clean again here between the build and the test?

README.md
```diff
@@ -20,14 +20,28 @@ Add the following to your `WORKSPACE` file to add the necessary external depende
 
 ```python
 git_repository(
+    name = "io_bazel_rules_docker",
+    commit = "{HEAD}",
```
dlorenc: Why are we using {HEAD} here? This won't work when people copy/paste.

examples/hello-grpc/proto/simple.proto
```diff
@@ -0,0 +1,15 @@
...
+  rpc Foo (FooRequest) returns (FooReply) {}
+}
+
+message FooRequest {
+  string name = 1;
+}
+
+message FooReply {
+  string message = 1;
+}
```
dlorenc: nit: newline

k8s/resolver.py
```diff
@@ -0,0 +1,118 @@
...
+parser = argparse.ArgumentParser(
+    description='Resolve image references to digests.')
+
+parser.add_argument(
+  '--template', action='store',
+  help='The template file to resolve.')
+
+
+def Resolve(input, tag_to_digest):
+  """Translate tag references within the input yaml into digests."""
```
dlorenc: So the idea is that *all* image references in a yaml get resolved into a digest? Is there a way to specify just a subset of them to resolve? I guess it's not a big deal, but I'm thinking of cases where users specify sidecar containers maintained by other teams/projects.

k8s/resolver.py
```diff
@@ -0,0 +1,118 @@
...
+  transport = transport_pool.Http(httplib2.Http, size=_THREADS)
+
+  overrides = {}
+  # TODO(mattmoor): Support publishing images and using the
+  # published digest to override resolution (to avoid races).
+
+  with open(args.template, 'r') as f:
+    inputs = f.read()
+
+  print _DOCUMENT_DELIMITER.join([
```
dlorenc: print function vs. statement?

.travis.yml
mattmoor: Mostly just that I started from a copy/paste of the line above. :)

I'm happy to change if you feel strongly.

README.md
mattmoor: I added this when going through the OSS process because I didn't have a commit yet.  The "can't copy/paste" is somewhat by design for now, I think that once we cut a first release we should update this to reference releases.

k8s/resolver.py
mattmoor: Yeah, it resolves everything.  Once I add `images={}` it may make sense to also add an option that will tell this to *only* resolve published images, but I think my preference is just that: opt out.

k8s/resolver.py
mattmoor: Old habit.  Don't tell Doug :)

examples/hello-grpc/proto/simple.proto
mattmoor: Done.

mattmoor:
Should be RFAL.  thanks for the feedback.

k8s/BUILD
```diff
@@ -14,3 +14,37 @@
...
+    data = [
+        "//examples/hello-grpc/cc/server:server.tar",
+    ],
+    deps = [
+        ":resolver_lib",
+        "@mock//:mock",
+    ],
+)
+
+par_binary(
```
dlorenc: It would probably make sense for this to be a standalone tool packaged as part of the containerregistry releases. But this LGTM for now.

PR #8, opened 2017-09-07
Add support for the `images={}` attribute.

mattmoor:
When the `images={}` attribute is specified resolution will publish the collection of images, and substitute the exact digest of the published image as part of resolution (avoids race condition).

### DIFF
--- a/k8s/object.bzl
+++ b/k8s/object.bzl
@@ -47,7 +47,6 @@ def _impl(ctx):
     # Each images entry results in a single --image_spec argument.
     # As part of this walk, we also collect all of the image's input files
     # to include as runfiles, so they are accessible to be pushed.
-    index = 0
     for tag in ctx.attr.images:
       target = ctx.attr.images[tag]
       image = _get_layers(ctx, image_target_dict[target], image_files_dict[target])
@@ -72,7 +71,6 @@ def _impl(ctx):
         "%s=%s" % (k, v)
         for (k, v) in image_spec.items()
       ])]
-      index += 1
 
   ctx.action(
       command = """cat > {resolve_script} <<"EOF"
--- a/k8s/object.bzl
+++ b/k8s/object.bzl
@@ -13,6 +13,16 @@
 # limitations under the License.
 """An implementation of k8s_object for interacting with an object of kind."""
 
+load(
+    "@io_bazel_rules_docker//docker:layers.bzl",
+    _get_layers = "get_from_target",
+    _layer_tools = "tools",
+)
+load(
+    "@io_bazel_rules_docker//docker:label.bzl",
+    _string_to_label = "string_to_label",
+)
+
 def _impl(ctx):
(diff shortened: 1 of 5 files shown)


mattmoor:
@dlorenc It is notable that this won't work with `minikube` unless the user's roundtripping through a real registry.  

(thinking out loud)
That said, we can probably put together an alternate implementation of `Publish()` that supports `docker load`ing the image into `minikube`'s docker daemon.  If we put this in a little library in `rules_docker` then we could also use it to implement [this](https://github.com/bazelbuild/rules_docker/issues/72), which is a nice two-for-one.

WDYT?

k8s/object.bzl
```diff
@@ -23,25 +33,70 @@ def _impl(ctx):
...
+        ctx.attr.image_targets, ctx.attr.image_target_strings)
+    image_files_dict = _string_to_label(
+        ctx.files.image_targets, ctx.attr.image_target_strings)
+
+    # Walk the collection of images passed and for each key/value pair
+    # collect the parts to pass to the resolver as --image_spec arguments.
+    # Each images entry results in a single --image_spec argument.
+    # As part of this walk, we also collect all of the image's input files
+    # to include as runfiles, so they are accessible to be pushed.
+    index = 0
```
dlorenc: Is index used anywhere?

mattmoor: Yeah, where I increment it :)  Removed.

mattmoor:
please test this

mattmoor:
test this please

rohansingh:
> That said, we can probably put together an alternate implementation of Publish() that supports docker loading the image into minikube's docker daemon.

@mattmoor I've put together some rules and a macro for a repo that basically does this. Notes for anyone that wants to take a stab on a real implementation:

1. The difficult part was deciding how to tag the images locally, since locally-loaded Docker images can't be referenced by digest. I ended up just tagging the images with their digest value.

2. I didn't want to tackle [rewriting the incremental loader](https://github.com/bazelbuild/rules_docker/issues/72) from `rules_docker` in Python. Instead, I just created a separate rule that wraps `incremental_load` from `@io_bazel_rules_docker//container:layer_tools.bzl` to load the image, and then wraps `k8s_object` to do Kube stuff.

3. Because of the above, I didn't change `Publish()` to load the image into the Docker daemon. Instead, the resolver just accepts a `--no_upload` flag. When passed, it does not upload the image. Instead, it resolves the image name using the tag strategy described above.

Obviously this is a bit of a hacky solution, but it seems to work quite reliably. Having the ability to deploy manifests locally like this is really handy.

mattmoor:
FWIW I solved this in `ko` by simply tagging the image with the digest, see: https://github.com/google/go-containerregistry/blob/master/cmd/ko/README.md

mattmoor:
Note: my bias towards not using `:latest` was that it doesn't work well with K8s default pull policy.

PR #14, opened 2017-09-11
Add support for special actions

mattmoor:


k8s/object.bzl
```diff
@@ -96,9 +96,67 @@ EOF""".format(
...
+./{resolve_script} | \
+  kubectl --cluster="{cluster}" --namespace="{namespace}" replace -f -
+EOF""")
+
+def _describe_impl(ctx):
+  return _common_impl(ctx, """cat > {script} <<"EOF"
+#!/bin/bash -e
+set -o pipefail
+kubectl --cluster="{cluster}" --namespace="{namespace}" describe {kind} \
+  "$(kubectl create --dry-run -f "{unresolved}" | cut -d'"' -f 2)"
```
dlorenc: What's the create --dry-run for?

k8s/object.bzl
```diff
@@ -96,9 +96,67 @@ EOF""".format(
...
+    kwargs["resolve_script"] = ctx.executable.resolved.short_path
+    files += [ctx.executable.resolved]
+    files += list(ctx.attr.resolved.default_runfiles.files)
+
+  if hasattr(ctx.files, "unresolved"):
+    kwargs["unresolved"] = ctx.files.unresolved[0].short_path
+    files += ctx.files.unresolved
+
+  ctx.action(
+      command = template.format(**kwargs),
```
dlorenc: Would it make more sense to use something like: https://docs.bazel.build/versions/master/skylark/lib/actions.html#expand_template

To generate the templated file?

k8s/object.bzl
mattmoor: I hoisted this into a variable name.  Essentially I'm abusing this to extract the resource name.  I'm open to suggestions :-/

k8s/object.bzl
mattmoor: I moved the string passed from each caller into a file, and now everything goes through `ctx.actions.expand_template` and shares a common implementation :)

mattmoor:
Should be RFAL

mattmoor:
Hmm, I thought I'd disabled concurrent jobs on Travis, but apparently not.  I'm going to close and re-open to confirm it is now properly configured.

PR #20, opened 2017-09-14
Rename the sample app, otherwise a Python sample is untennable.

mattmoor:


mattmoor:
This is a pretty straightforward, mostly scripted change:
```shell
git mv examples/hello-grpc/ examples/hellogrpc/
sed -i 's@examples/hello-grpc@examples/hellogrpc@g' $(find -type f | xargs grep examples/hello-grpc | cut -d':' -f 1 | uniq)
```

If testing passes, it should be RFAL.

PR #22, opened 2017-09-16
This adds a sample Python GRPC service.

mattmoor:
Currently this depends on [my PR](https://github.com/bazelbuild/rules_docker/pull/146) to fix an issue I'd run into with how `py_image` was laying out external dependencies.  Once this is merged and I've updated the PR to point at a commit on the main `rules_docker` repo, I'll remove the `[WIP]`, but otherwise this should be good to go.


mattmoor:
@dlorenc Following this example, it should be pretty reasonable to stand up a GRPC interface for doing FTL builds :)

mattmoor:
The PR to `rules_docker` is in and this PR has been updated, this is RFAL.

PR #34, opened 2017-09-21
Bugfix: Extracts short_path from tarball-File

DiTo04:
This removes the bug not allowing k8s_object() to work on images from docker_build() targets with a legacy layer.

I saw your pending change to move from .short_path and this will need updating as well.

And good job with rules_k8s! :)
//David, from Cinnober

### DIFF
--- a/k8s/object.bzl
+++ b/k8s/object.bzl
@@ -60,7 +60,7 @@ def _impl(ctx):
 
       image_spec = {"name": tag}
       if image.get("legacy"):
-        image_spec["tarball"] = image["legacy"]
+        image_spec["tarball"] = image["legacy"].short_path
         all_inputs += [image["legacy"]]
 
       blobsums = image.get("blobsum", [])


bazel-io:

DiTo04:
Here is a comment to trigger a new check on the CLA, as I have added @cinnober.com as an alternative adress.

mattmoor:
@DiTo04 thanks, this looks right.  I really wish Travis had a way for me to tell it to run tests with secrets :-/

mattmoor:
@DiTo04 I updated my PR to drop `.short_path` as well if you want to give it another shot.

Out of curiosity, do you know why you are running into the legacy tarball path?

There are some places it is unavoidable, e.g. coming from `docker save` like [this](https://github.com/bazelbuild/rules_docker/blob/master/docker/testdata/BUILD#L281).

There are also places it is avoidable, but an unnecessary `.tar` has been specified like [this](https://github.com/bazelbuild/rules_docker/blob/master/docker/testdata/BUILD#L288).

The TMI version is [here](https://github.com/bazelbuild/rules_docker/blob/master/intermediate-form.md), but suffice it to say that if you can avoid this path, things should be faster.

DiTo04:
I was specifying the  `docker_build` to use the `.tar`... I changed that and it works like a charm. Thank you for noticing! :)

mattmoor:
No problem, I'm glad it was so easy :)

PR #38, opened 2017-09-22
Python HTTP Example

mattmoor:
A broken Python HTTP example, to which I will point a rules_python issue.


mattmoor:
I updated the `Flask` dependency because the latest version of `Flask` has a `.whl`, but `itsdangerous` and `MarkupSafe` both still have problems. :(

mattmoor:
Alright, the Travis failure was because the image reference wasn't renamed in the same manner as I renamed things in the template when I added image chrooting.  Another PR makes this an error.

PR #45, opened 2017-09-29
Update jsonnet dependency

mattmoor:


WORKSPACE
```diff
@@ -156,14 +156,10 @@ load("@org_pubref_rules_protobuf//python:rules.bzl", "py_proto_repositories")
...
+git_repository(
     name = "io_bazel_rules_jsonnet",
-    sha256 = "e0f42a31f5b2feeeb5496d40cb0b7c5d3872bd7e0bdb44b04543cce96de321f8",
-    strip_prefix = "rules_jsonnet-0.0.3",
-    urls = [
-        "http://mirror.bazel.build/github.com/bazelbuild/rules_jsonnet/archive/0.0.3.tar.gz",
-        "https://github.com/bazelbuild/rules_jsonnet/archive/0.0.3.tar.gz",
-    ],
+    commit = "09ec18db5b9ad3129810f5f0ccc86363a8bfb6be",
+    remote = "https://github.com/bazelbuild/rules_jsonnet.git",
```
damienmg: Why not keep the tarball with http_archive for speed?

mattmoor: Because nobody re-released this change.

damienmg: but you can use a tarball of a commit directly, see https://bazel-review.googlesource.com/c/eclipse/+/17531 for instance

PR #48, opened 2017-09-29
This change adds preliminary support for k8s_objects.

mattmoor:
`k8s_objects` takes a list of K8s objects across which to take certain actions.

`k8s_objects` also supports nesting, so a `k8s_objects` can take another `k8s_objects`.

Fixes: https://github.com/bazelbuild/rules_k8s/issues/46

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ Travis CI | Bazel CI
 
 * [k8s_defaults](#k8s_defaults)
 * [k8s_object](#k8s_object)
+* [k8s_objects](#k8s_objects)
 
 ## Overview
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -117,6 +118,56 @@ k8s_deploy(
 )
 ```
 
+### Multi-Object Actions
+
+It is common practice in the Kubernetes world to have multiple objects that
+comprise an application.  There are two main ways that we support interacting
+with these kinds of objects.
+
+The first is to simply use a template file that contains your N objects
```
dlorenc: Is this basically a multi-document YAML file, or are we doing anything special here?

k8s/objects.bzl
```diff
@@ -0,0 +1,73 @@
...
+      substitutions = {
+          "%{resolve_statements}": ("\n" + ctx.attr.delimiter).join([
+              _runfiles(ctx, exe.files_to_run.executable)
+              for exe in ctx.attr.objects
+          ]),
+      },
+      output = ctx.outputs.executable,
+  )
+
+  runfiles = [obj.files_to_run.executable for obj in ctx.attr.objects]
```
dlorenc: Should we be using `depset`s here instead of lists?

README.md
```diff
@@ -117,6 +118,56 @@ k8s_deploy(
...
+The second is through the use of `k8s_objects`, which aggregates N `k8s_object`
+rules:
+```python
+# Note the plurality of "objects" here.
+load("@io_bazel_rules_k8s//k8s:objects.bzl", "k8s_objects")
+
+k8s_objects(
+   name = "deployments",
+   objects = [
+      ":foo-deployment",
```
dlorenc: Is there a way to use glob to specify a wildcard on a set of these? It might make a good example.

README.md
mattmoor: Just a multi-part yaml, nothing fancy.

README.md
mattmoor: In this context, they are targets, so not really.

You could use a glob to generate the `k8s_deploy` rules:
```python
[k8s_kind(
    name = "some-function-of-input",
    template = f,
) for f in glob(["*.kind.yaml"])]
```

I don't do this in Prow, but I do something similar.

k8s/objects.bzl
mattmoor: `ctx.runfiles(files = ...)` expects a `sequence of Files`, so `depset` really only helps with potential deduping, which I'm not super concerned about (I'd guess Bazel does this later anyways?).

PR #56, opened 2017-10-12
Update rules_protobuf

vladmos:
Starting from Bazel 0.8 the += operator applied to lists will mutate them instead of copying. This is an incompatible behavior and repostories should be fixed in advance.

To reproduce the future behavior one can use the following flag:

    bazel build --incompatible_list_plus_equals_inplace //...

This PR updates rules_protobuf so that the following commit is also included: https://github.com/pubref/rules_protobuf/commit/4cc90ab2b9f4d829b9706221d4167bc7fb3bd247

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -72,7 +72,7 @@ py_library(
 
 git_repository(
     name = "org_pubref_rules_protobuf",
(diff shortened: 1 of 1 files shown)


mattmoor:
Merging this into a branch off `master`, so that I can verify Travis passes before merging to `master`.

PR #63, opened 2017-10-20
Fix a bug that was stopping us from resolving tags we don't push.

mattmoor:


dlorenc:
Should we add a test-case to resolver_test.py?

mattmoor:
I added a rest to the e2e, since this code is hard to unit test.

mattmoor:
Test this please

PR #66, opened 2017-11-09
Add a pre-processor for templates to improve json support.

mtsgrd:
I've added a pre-processor for converting any user-supplied templates to yaml before stamping and resolving image tags. This is necessary for working with multi-resource json objects (arrays).

Is this the right way to go or did anyone have another solution in mind?

### DIFF
--- a/k8s/BUILD
+++ b/k8s/BUILD
@@ -33,6 +33,7 @@ py_library(
     deps = [
         "@containerregistry//:containerregistry",
         "@yaml//:yaml",
+        "@simplejson//:simplejson",
     ],
 )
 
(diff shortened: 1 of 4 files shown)


bazel-io:

mattmoor:
tl;dr The rules should already support this.

YAML v2 is a superset of JSON, so we just parse it as yaml.  Before we do we split the document on `---`, and then join on it when we're done.  Is that why you closed this?

mtsgrd:
I closed the pull request because I'm on a new computer and forgot to configure my git email. While fixing the commit, it dawned on me that I should be using multiple outputs combined with giving `kubectl` a directory as input. Does this sound right?

mtsgrd:
And now that I've taken a closer look I realize `k8s_object` only takes single input template. I can certainly have one per output json file, which is useful since it gives me finer control.. but it would be great to hear from you if that's what is intended?

PR #67, opened 2017-11-14
Update docs.

ensonic:
Update the docs for the docker_rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ git_repository(
 )
 
 load(
-  "@io_bazel_rules_docker//docker:docker.bzl",
-  "docker_repositories",
+  "@io_bazel_rules_docker//container:container.bzl",
+  container_repositories = "repositories",
 )
 
-docker_repositories()
+container_repositories()
 
 # This requires rules_docker to be fully instantiated before
 # it is pulled in.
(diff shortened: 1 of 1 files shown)


bazel-io:

README.md
```diff
@@ -28,10 +28,10 @@ git_repository(
 
 load(
   "@io_bazel_rules_docker//docker:docker.bzl",
```
mattmoor: We should load this from `container:container.bzl` if we're going to use this name (which we should!).

ensonic: Well spotted. Thanks and fixed.

README.md
```diff
@@ -203,7 +203,6 @@ the following:
 ```shell
 $ cat .bazelrc
 build --workspace_status_command=./print-workspace-status.sh
-run --workspace_status_command=./print-workspace-status.sh
```
mattmoor: Why did you remove this?

ensonic: See commit message:
Simplify .bazelrc example.  …
According to
https://docs.bazel.build/versions/master/command-line-reference.html#run
run inherits from build and therefore we can omit the entry for run.

mattmoor:
Cool thanks.  Travis only runs properly from PRs coming from branches on the main repo (b/c of secrets), and I'm just going to skip Bazel CI because this is README only.

Issue #69, opened 2017-11-14
Remove unchanged/useless output from k8s_object


ensonic:
If one does:
```
k8s_robco_object(
  name = "config",
  kind = "configmap",
  template = ":config.yaml",
)
```
you will get a an error (see subject). One can work around it by using `name="x-config"` but that is ugly.

mattmoor:
I changed the title.  Currently this does nothing, so we should just remove it.

ensonic:
I am not sure it is useless. What is the reason for not outputting the expanded yaml file when running build?

mattmoor:
What comes out during the build isn't expanded (pushing wouldn't be hermetic).

Right now it is literally copying the input to the output :)

ensonic:
Yes, the current output is not useful. It would only be useful if it would output a yaml with images resolved. It should be hermetic, since the image would point to the digest.

mattmoor:
Right, and we can only publish the image and produce a resolved template on `bazel run` since that is inherently not hermetic.

mattmoor:
Just to be clear, our hermetic references are to different things :)

What I'm saying is non-hermetic is publishing an image, since it talks to the registry service and accesses things outside of the sandbox like your credential keyring.

The result of `bazel run` (the resolved output) is what I'd call pseudo-hermetic because all of the external references are now content-addresses.  So while it isn't closed per-se, the only way you can successfully deploy it is if get exactly what you intended.

Issue #68, opened 2017-11-14
Better docs for image substitutions


ensonic:
Lets assume I have a deployment:
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: web-frontend
spec:
  replicas: 1
  template:
    spec:
      - name: web-frontend
        image: XXXX
```
and a build rule:
```
k8s_object(
  name = "x-deployment",
  kind = "deployment",
  template = ":deployment.yaml",
  images = {
    "XXXX": "//src/docker/web_frontend:web-frontend",
  }
)
```

I get:
```
INFO: Running command line: bazel-bin/src/kubernetes/web-frontend/x-deployment
Traceback (most recent call last):
  File "/usr/local/google/home/ensonic/.cache/bazel/_bazel_ensonic/execroot/__main__/bazel-out/local-fastbuild/bin/src/kubernetes/web-frontend/x-deployment.runfiles/__main__/../io_bazel_rules_k8s/k8s/resolver.py", line 180, in <module>
    main()
  File "/usr/local/google/home/ensonic/.cache/bazel/_bazel_ensonic/execroot/__main__/bazel-out/local-fastbuild/bin/src/kubernetes/web-frontend/x-deployment.runfiles/__main__/../io_bazel_rules_k8s/k8s/resolver.py", line 166, in main
    (tag, digest) = Publish(transport, args.image_chroot, **kwargs)
  File "/usr/local/google/home/ensonic/.cache/bazel/_bazel_ensonic/execroot/__main__/bazel-out/local-fastbuild/bin/src/kubernetes/web-frontend/x-deployment.runfiles/__main__/../io_bazel_rules_k8s/k8s/resolver.py", line 136, in Publish
    name_to_replace = docker_name.Tag(name)
  File "/usr/local/google/home/ensonic/.cache/bazel/_bazel_ensonic/execroot/__main__/bazel-out/local-fastbuild/bin/src/kubernetes/web-frontend/x-deployment.runfiles/containerregistry/client/docker_name_.py", line 196, in __init__
    _check_tag(self._tag)
  File "/usr/local/google/home/ensonic/.cache/bazel/_bazel_ensonic/execroot/__main__/bazel-out/local-fastbuild/bin/src/kubernetes/web-frontend/x-deployment.runfiles/containerregistry/client/docker_name_.py", line 80, in _check_tag
    _check_element('tag', tag, _TAG_CHARS, 1, 127)
[...]

PR #71, opened 2017-11-16
Address a few pieces of feedback from @ensonic

mattmoor:
Removes some of the friction from: https://github.com/bazelbuild/rules_k8s/issues/68
Fixes: https://github.com/bazelbuild/rules_k8s/issues/69

### DIFF
--- a/examples/hellohttp/BUILD
+++ b/examples/hellohttp/BUILD
@@ -53,7 +53,7 @@ k8s_deploy(
     name = "error-on-run",
     images = {
         # Simulate a typo
-        "us.gcr.io/not-my-porject/hello-http:staging": "//examples/hellohttp/go:server",
+        "hello-http_image": "//examples/hellohttp/go:server",
     },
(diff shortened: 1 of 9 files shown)


mattmoor:
@ensonic This should enable you to use a much more relaxed set of things as the keys in the images dictionary.  It isn't completely free-form because it still becomes a suffix of the image name, but I also disable strictness, so things like tags can be inferred.

If you could take this for a spin, and LMK what you think I'd appreciate it :)

mattmoor:
@ensonic Were you able to make this work?

mattmoor:
@ensonic LOL, yeah this was one of my original motivations for sticking with strict image names.  I'm not sure how to avoid it without some level of schema-awareness.

ensonic:
Yeah, it is tricky. We should merge this. I've filed a new ticked to discuss the expansion:
https://github.com/bazelbuild/rules_k8s/issues/83

PR #95, opened 2018-01-11
Make resolver a public property

stevewolter:
The custom resolver mattmoor@ mentioned seems to be a very workable solution for endpoint resolving (#88). I've made a PoC implementation in http://shortn/_0SyJs3BlK7 (Google-internal link) that uses a custom resolver to resolve endpoint versions after Docker digests were resolved. Could you PTAL?

PR #96, opened 2018-01-11
Make resolver a public property (#95)

mattmoor:
* Make resolver a public property

* Review comments


mattmoor:
@stevewolter FYI

stevewolter:
Confirming that I'm ok with all commits from #95 and #96 being contributed to this project under CLA.

PR #114, opened 2018-02-21
describe: get the object kind from the config

### LINKED ISSUE #113: .describe doesn't work if k8s template creates multiple resources
I have k8s jsonnet templates that output two resources: a deployment and a service.

That works fine for `. apply` actions, but the `.describe` ones seem to get confused:

```
$ bazel run //examples/hellohttp/go/server:go-hellohttp_deploy.describe
INFO: Analysed target //examples/hellohttp/go/server:go-hellohttp_deploy.describe (157 packages loaded).
INFO: Found 1 target...
Target //examples/hellohttp/go/server:go-hellohttp_deploy.describe up-to-date:
  bazel-bin/examples/hellohttp/go/server/go-hellohttp_deploy.describe
INFO: Elapsed time: 13.445s, Critical Path: 0.53s
INFO: Build completed successfully, 1 total action

INFO: Running command line: bazel-bin/examples/hellohttp/go/server/go-hellohttp_deploy.describe
Error from server (NotFound): deployments.extensions "go-hellohttp-app-deployment\ngo-hellohttp-svc" not found
ERROR: Non-zero return code '1' from command: Process exited with status 1
```

The underlying reason is that the `kubectl create --dry-run` returns multiple resources but the describe script expects only one.

```
$ kubectl create --dry-run -f "${RUNFILES}/__main__/test_cd_deploy.yaml" | cut -d'"' -f 2
test-cd-app-deployment
test-cd-svc
```

Code that consumes this:

``` 
RESOURCE_NAME=$(kubectl create --dry-run -f "${RUNFILES}/__main__/test_cd_deploy.yaml" | cut -d'"' -f 2)

kubectl --cluster="fooo" --namespace="$(cat ${RUNFILES}/__main__/test_cd_deploy.describe.namespace-name)" \
    describe deployment "${RESOURCE_NAME}"
```

Should we just replace [this](https://github.com/bazelbuild/rules_k8s/blob/d6bca8104db9890d66e0b0aa1a3c558d985ed8d3/k8s/describe.sh.tpl#L26) with a for loop?

nictuku:
This should fix #113 by showing all resources found in the config.

Does this mean we can actually remove the "kind" attribute from the bazel rules and instead always read it from the config directly - like we're doing here?

I believe that with this change, we don't need `kind` anywhere else, or do we?

### DIFF
--- a/k8s/describe.sh.tpl
+++ b/k8s/describe.sh.tpl
@@ -23,7 +23,9 @@ function guess_runfiles() {
 
 RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
 
-RESOURCE_NAME=$(kubectl create --dry-run -f "%{unresolved}" | cut -d'"' -f 2)
-
-kubectl --cluster="%{cluster}" %{namespace_arg} describe %{kind} \
-  "${RESOURCE_NAME}"
+kubectl create --dry-run -f "%{unresolved}" | tr -d '"' | \
+  while IFS=" " read -r KIND RESOURCE_NAME remainder
+  do
+    kubectl --cluster="%{cluster}" %{namespace_arg} describe "${KIND}" \
+      "${RESOURCE_NAME}"
+  done


bazel-io:

nictuku:
Heads-up that the CI test failure is expected. The travis CI is failing only because  it refuses to run the tests when the PR is coming from an untrusted repo (which is the case).

I assume admins should be able to inspect the code and trigger a manual CI if they don't think my code will grab their credentials :).

mattmoor:
@nictuku I wish Travis were that smart.  Sadly I need to merge this into a branch on the main repo and reopen the PR to get things to run properly.  You should rebase before I do that :)

You are right that `kind` should no longer be needed (hooray!).  We should switch the assorted rules that take it to print a warning if it's specified (noting that it will become an error in the future).

Other things this enables:
1. We no longer need to predicate `.describe` on the presence of `kind`
1. `k8s_objects` can now work on `.describe` as well (since I think it should always be there)

nictuku:
OK from what you're saying the next step is for me to rebase. I'll do it ASAP. Thanks!

fejta:
Do you have an ETA for rebasing? Please do and update this PR.

Please close the PR if it is no longer needed and/or you no longer plan to work on it.

nictuku:
I don't need this at the moment so I'll just close it.

PR #118, opened 2018-03-02
Fix file reference to rules_docker

ash2k:
The file was renamed in https://github.com/bazelbuild/rules_docker/pull/279.


ash2k:
CLA should be fine now, please check.

PR #130, opened 2018-05-03
Fix k8s_object to work with latest rules_docker

### LINKED ISSUE #125: Depending on latest rules_docker breaks example
Not sure if it is a bug here or at rules docker, but depending on 

```
git_repository(
    name = "io_bazel_rules_docker",
    commit = "4d8ec6570a5313fb0128e2354f2bc4323685282a", # latest commit as of 27/03/2018
    remote = "https://github.com/bazelbuild/rules_docker.git",
)
```

breaks `bazel run //examples/todocontroller/py:controller-deployment`

The relevant PR for that commit is https://github.com/bazelbuild/rules_docker/pull/340

ixdy:
I don't really understand this change that well; I'm just mirroring the refactoring in https://github.com/bazelbuild/rules_docker/pull/340/commits/7228cb2764c519408478c70c6865090bdb856be7.

It seems to build, so maybe it's correct? I'm not sure if there's semantic meaning I'm missing, though.

Fixes #125

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -15,7 +15,7 @@ workspace(name = "io_bazel_rules_k8s")
 
 git_repository(
     name = "io_bazel_rules_docker",
-    commit = "27c94dec66c3c9fdb478c33994471c5bfc15b6eb",
+    commit = "452878d665648ada0aaf816931611fdd9c683a97",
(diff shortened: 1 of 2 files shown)


ixdy:
post-kubecon ping :)

PR #132, opened 2018-05-16
Add wait_for_rollout attr

ben-ng:
Required to support the deployment of more complex applications, where some resource types depend on others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,6 +473,15 @@ A rule for interacting with Kubernetes objects.
         <p>The repository under which to actually publish Docker images.</p>
       </td>
     </tr>
+    <tr>
+      <td><code>wait_for_rollout</code></td>
+      <td>
+        <p><code>bool, optional</code></p>
+        <p>Whether or not to wait for the rollout of a Deployment or StatefulSet. No effect if used on any other resource type.</p>
+        <p><b>If this is omitted, it will default to
+          <code>False</code>.</b></p>
+      </td>
+    </tr>
   </tbody>
 </table>
 
--- a/examples/hellogrpc/cc/server/BUILD
+++ b/examples/hellogrpc/cc/server/BUILD
@@ -31,4 +31,5 @@ k8s_deploy(
         "us.gcr.io/not-my-project/hello-grpc:staging": ":server",
     },
     template = "//examples/hellogrpc:deployment.yaml",
+    wait_for_rollout = True,
 )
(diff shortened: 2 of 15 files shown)


mattmoor:
I'm fine with the argument pass-thru (do you want to split off a PR to get this in, while we discuss the rest?).

I'm opposed to the `wait_for_rollout` parts of this change as-is, I'll try to elaborate on why.
1. I don't like doing this in bash. I'd say use the `kubernetes` python package to write a little tool, but this significantly complicates things in other ways that are undesireable. For instance, I think this logic would probably be faster and incur less load on the K8s master if it used watch. I get that this is somewhat the pot calling the kettle black (re: bash), but unfortunately there's a lot of functionality locked up in `kubectl` and not the libraries.
1. I don't like having 3x copies of the same logic
1. I don't like how this complicates the argument pass-thru (do those same argument belong on the later kubectl commands?)
1. I'm a little bothered by the schema-awareness this starts to add (we've largely stayed schema agnostic, and it's helped us support arbitrary CRDs)
1. I don't like that this introduces orchestration to the evaluation of the rules. Generally `bazel run :foo.bar` is comparable to `bazel run :foo | kubectl -n ... -c ... bar -f -`.  This is my goto recommendation for how folks use this for release management to decouple evaluation from instantiation, but the way this is solved outside of `kubectl` worries me.

I don't mean for this to sound like a rant (apologies if it comes across that way), I'm just thinking out loud as I page context back in.  We have a similar problem using these rules with Istio (and a few other CRDs) where:
1. A CRD is created,
2. An instance of that CRD is created, and the API server hasn't lit up the new routes yet (boo)

Thinking more about this, I think a split/factoring I could get behind would be to create a wrapper like `k8s_objects` that wraps `k8s_object[s]`-style rules.  It would go something like this:

```python
k8s_object(
    name = "foo",
    ...
)

# Live in: contrib/apps/something.bzl (?)
wait_for_rollout(
    name = "bar",
    object = "foo",
)

k8s_objects(
    name = "blah",
    objects = [
    ...
    ":bar",
    ...
    ],
)
```

This doesn't address all of my concerns, but it moves it out of the base primitives to a place where we can call out caveats (and I'm inclined to be less of a purist ;-D).  WDYT?

fejta:
Please address comments and update PR and/or close if this is no longer needed. Thanks!

ben-ng:
I'll have time to circle back and update this pull request in a few weeks. We're in the middle of building applications using our fork of rules_k8s, and I don't have the bandwidth to change two things at once.

fejta:
Awesome, looking forward to seeing the update once you have time to do so.

fejta:
/override continuous-integration/travis-ci/pr
Replaced with `pull-rules-k8s-e2e`

fejta:
/override continuous-integration/travis-ci/pr
Replaced with pull-rules-k8s-e2e

fejta:
Please reopen when you have time to work on this!
/close

Issue #135, opened 2018-05-22
Proposal: Allow to specify kubectl context instead of cluster


burdiyan:
It may be useful to allow to specify context for kubectl instead of cluster. And actually current docs are a little bit misleading, because it says "This should match the cluster name as it would appear in `kubectl config current-context`" but this command doesn't output anything related to cluster, only name of the context.

In our use case of GKE, we use `get-credentials` from `gcloud` to authenticate in the cluster, and then we use `kubectl config rename-context` to rename the complicated name to our standards. But we don't touch cluster names. So that's making things a little bit more complicated.

jasongwartz:
FYI, I'm going to work on a PR to add the context parameter, so it would be possible to use either context or cluster or both.

PR #138, opened 2018-05-30
Adds 'context' parameter to k8s_object

jasongwartz:
Apologies for opening this a second time. I had to rebase in order to get the right email address in the commit (for the CLA).

Related: #135

This PR introduces a new parameter to k8s_object, in order to manually specify the 'context' from the local kubeconfig file.
This is necessary in addition to the 'cluster' parameter in cases where the bazel rule should be applied into a 'context' which is configured on the local machine, but not currently selected.

For example: in our use case, we have a macro which generates and runs k8s_object rules for all of our deploy environments and clusters (dev, prod, multiple clusters) - which means it needs to run apply against different 'contexts' consecutively without changing the kubectl context.

cc @Globegitter


jasongwartz:
Ok, NOW it should confirmed signed.

fejta:
Does your `kubectl config get-contexts` differ from `kubectl config get-clusters`?

k8s/object.bzl
```diff
@@ -144,6 +144,19 @@ def _common_impl(ctx):
     cluster_arg = "$(cat %s)" % _runfiles(ctx, cluster_file)
     files += [cluster_file]
 
+  
+  # If the 'context' parameter is not set by the caller,
+  #     this value becomes an empty string, and kubectl
+  #     will be run like `kubectl --context= ...` In this
```
fejta: Would want this to be `kubectl --context={cluster}` instead?

jasongwartz:
@fejta Well - in my personal `.kube/config`, the context names are the same as the cluster names, but I have, in the past, set 'friendly' context names for clusters with very similar names (in combination with using https://github.com/ahmetb/kubectx).

In fact, the first tests of this PR didn't add an extra bazel parameter and just modified the `kubectl` command adding `--context={cluster}`, but I decided it was safer to accommodate more generalised use-cases, hence being able to set `{cluster}` and `{context}` independently.

I hope I'm understanding your comment correctly!

k8s/object.bzl
fejta: No it seems like `--context=` is better

jasongwartz:
@fejta  Great, thanks for reviewing!

I'm not sure what else needs to be done, or if an owner like you just needs to press the button.

jasongwartz:
Ok, no problem!

PR #144, opened 2018-06-20
Use the correct rule for k8s_service in hellogrpc service.

bweston92:
(this may be wrong but I can't see k8s_service being used anywhere and a service isn't a deployment)


nlopezgi:
/override cla/google
(see above)

Issue #146, opened 2018-06-20
Pass arbitrary arguments to kubectl


mariusgrigoriu:
#126 and #145 are requesting additional arguments to kubectl, and I suspect there will be more such issues in the future as the user base and the variety of scenarios grow. kubectl has many flags and it seems that we will be perpetually behind. Should we have a way to pass arbitrary arguments to kubectl?

chrislovecnm:
https://github.com/bazelbuild/rules_k8s/pull/224

Is merged. Closing

PR #148, opened 2018-06-22
Support kubectl --user

mariusgrigoriu:
#### Why this PR is needed

If a user running rules_k8s is setting `cluster` to a value that is not the same as the cluster in their current context, then there is a chance that they also need to set a different user.

#### Approached used

I naively searched for the `cluster` attribute and added matching code for a `user` attribute.


fejta:
Would https://github.com/bazelbuild/rules_k8s/pull/138 address this need instead?

mariusgrigoriu:
@fejta #138 could address the need with some adaptation on our end to set contexts -- we're overriding namespace, user, and cluster, but don't have a context setup for this specific case. But rules_k8s already supports two out of the three. 

I think this brings me back to the problem statement from #146. Kubectl has many flags and, eventually, we may need access to more and more of them. Seems like replicating all the kubectl flags here is unsustainable.

mariusgrigoriu:
@fejta I think build is broken due to something outside of this PR. Can you confirm?

fejta:
/ok-to-test

> I think build is broken

The job is passing now! 

We're replacing the Travis CI job with its prow equivalent. Your PR won the prize for debugging things.

Forgot it would send a bunch of spam whenever it failed. Apologies and thanks!

mariusgrigoriu:
Still looks like some external issue is blocking the build

PR #156, opened 2018-07-02
Create gcloud-bazel image for running e2e tests

fejta:
/assign @erain

Next will create a prow job which uses this image and runs `test-e2e.sh`


images/gcloud-bazel/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Ubuntu with bazel, gcloud and its dependencies preinstalled.
+
+FROM ubuntu:16.04
```
erain: Could we use a google-maintained base image? 

https://console.cloud.google.com/launcher/details/google/ubuntu16_04

images/gcloud-bazel/BUILD.bazel
```diff
@@ -0,0 +1,2 @@
+# TODO(fejta): replace Dockerfile with an equivalent bazel rule. This seems promising:
+#   https://github.com/GoogleContainerTools/base-images-docker/blob/master/package_managers/bootstrap_image.bzl
```
erain: Yes I think (for TODO) package manager is the way to go. 

@xingao267 has done quite many works around package manager.

images/gcloud-bazel/Makefile
```diff
@@ -0,0 +1,30 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
erain: Using a Makefile is not ideal to me...

We usually use a [cloudbuild.yaml](https://cloud.google.com/container-builder/docs/create-custom-build-steps) to build a container even for only one step (build from a Dockerfile). In this way we can ensure all the containers are build in a controlled environment with [Google Cloud Container Builder](https://cloud.google.com/container-builder/docs/)

images/gcloud-bazel/BUILD.bazel
```diff
@@ -0,0 +1,2 @@
+# TODO(fejta): replace Dockerfile with an equivalent bazel rule. This seems promising:
```
xingao267: License header

images/gcloud-bazel/BUILD.bazel
xingao267: Yes. Using package manager is promising. We've been using that to install java, wget, git, bazel, docker, and etc in our containers. But we haven't tried gcloud yet (it's on our list), and we are not sure if extra work will be needed to make that work (it could be just as simple as what we did below for Bazel).

I think we can use Dockerfile for now. I'll look into installing gcloud, and once that works. We can change this Dockerfile to bazel rules.

For your interest, we use rules here https://github.com/bazelbuild/bazel-toolchains/blob/master/container/rules/docker_toolchains.bzl to create containers. These rules wrap rules in https://github.com/GoogleContainerTools/base-images-docker/blob/master/package_managers. Example usage of the rules are:

https://github.com/bazelbuild/bazel-toolchains/blob/master/container/ubuntu16_04/builders/bazel/BUILD#L23
https://github.com/bazelbuild/bazel-toolchains/blob/master/container/ubuntu16_04/layers/bazel/BUILD

images/gcloud-bazel/BUILD.bazel
fejta: Done

images/gcloud-bazel/BUILD.bazel
fejta: Ack

I'll circle back later and try and figure out how to use these rules.

images/gcloud-bazel/Dockerfile
fejta: Done

images/gcloud-bazel/Makefile
fejta: Done

The Makefile is necessary to get the tags I want, but it now uses gcb to push builds.

PR #157, opened 2018-07-06
switch from native.git_repository to external git_repository

excavador:
switch from native.git_repository to external git_repository from '@bazel_tools//tools/build_defs/repo:git.bzl' (#140)


excavador:
So, can we merge this PR?

PR #158, opened 2018-07-07
pass visibility into k8s commands

dmitrig01:
Fantastic tool! I ran into an issue creating k8s objects across multiple packages: I would like to run k8s_objects to aggregate objects created across multiple different packages. However, this isn't possible as they need to be marked as visibility:public, and this doesn't get passed to the underlying rule. This PR fixes that.


fejta:
/skip continuous-integration/travis-ci/pr
Replaced by `pull-rules-k8s-e2e`

fejta:
> /skip

Oh that isn't how this plugin works right now...
https://github.com/kubernetes/test-infra/blob/master/prow/plugins/skip/skip.go#L36

fejta:
@erain do you need to approve this to get the pending results to run?

PR #162, opened 2018-07-16
Add --allow_unused_images resolver flag (#161)

### LINKED ISSUE #161: Add --allow_unused_images resolver flag
# Background

Some macros will instantiate `k8s_object(...)` multiple times with different SKUs. When an image is only used by some SKUs and not others, the default resolver dies when it doesn't find the image in a particular SKU's template:

https://github.com/bazelbuild/rules_k8s/blob/4348f8e28b70cf3aff7ca8e008e8dc7ac49bad92/k8s/resolver.py#L190-L195

For cases like this, it would be nice to be able to relax the unused image error.

# Proposal

Add an `--allow_unused_images` flag to `resolver.py`. Add a `resolver_args` parameter to `k8s_object` that allows arguments to be passed directly to the resolver. This would be useful for custom resolvers as well.

calder:
This fixes #161.


calder:
@fejta @nlopezgi: Friendly ping. :-)

fejta:
/assign
Replied on the issue

calder:
@fejta: Thanks! Rereplied. :-)

calder:
@fejta @nlopezgi: Friendly ping!

fejta:
/override cla/google
(false-positive from my updating the branch)

PR #143, opened 2018-06-18
Propagate arguments for implicit attributes to instantiated rules (#137)

mattmoor:
(this runs Travis)

PR #137, opened 2018-05-30
Propagate implicitly added attributes/arguments to instantiated rules

ash2k:
Fixes #123.

PR #165, opened 2018-07-24
Rebase: Propagate arguments for implicit attributes to instantiated rules

ash2k:
Rebased version of https://github.com/bazelbuild/rules_k8s/pull/137 and https://github.com/bazelbuild/rules_k8s/pull/143.

### DIFF
--- a/k8s/object.bzl
+++ b/k8s/object.bzl
@@ -402,6 +402,17 @@ _k8s_object_delete = rule(
     implementation = _common_impl,
 )
 
+# See "attrs" parameter at https://docs.bazel.build/versions/master/skylark/lib/globals.html#parameters-26
+_implicit_attrs = ["visibility", "deprecation", "tags", "testonly", "features"]
+
+def _implicit_args_as_dict(**kwargs):
+    implicit_args = {}
+    for attr_name in _implicit_attrs:
+      if attr_name in kwargs:
+        implicit_args[attr_name] = kwargs[attr_name]
+
+    return implicit_args
+
 def k8s_object(name, **kwargs):
   """Interact with a K8s object.
   Args:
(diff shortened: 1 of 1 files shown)


k8s/object.bzl
```diff
@@ -402,6 +402,17 @@ _k8s_object_delete = rule(
     implementation = _common_impl,
 )
 
+# See "attrs" parameter at https://docs.bazel.build/versions/master/skylark/lib/globals.html#parameters-26
+_implicit_attrs = ["visibility", "deprecation", "tags", "testonly", "features"]
```
fejta: Would you be willing to use each of these in a test somewhere?

fejta: Request here would basically be to update something in the repo to include a testonly = 1, visibility = //foo:__subpackages__, etc so that when we bazel build //... something will fail if we lose support for this

fejta:
Is this alive? Can you please update some of the tests in this repo to use at least one of the "visibility", "deprecation", "tags", "testonly", "features"?

ash2k:
I spent a few hours trying to add a test but I could not figure it out. I don't have much bandwidth to finish this PR, sorry.

fejta:
Shall we close then? Testing is an important part of Google's development practices.

Alternatively where did you get stuck? The goal here is to create a k8s_object somewhere in this repo that has a `visibility` attribute. It doesn't really matter where the rule lives. https://github.com/bazelbuild/rules_k8s/blob/c861e4ea5a0b34e17fb682f60fa78a9c85050519/examples/todocontroller/BUILD is one place to create one.

One potential problem is that all the existing rules use `k8s_defaults`, and it doesn't look like this PR handles that scenario yet. Can you add the functionality there as well? https://github.com/bazelbuild/rules_k8s/blob/master/k8s/with-defaults.bzl

Globegitter:
This is something we have run into as well I should have time to add some tests for this - I will take this as a base and create a new PR in the next couple of days.

erain:
Thanks @Globegitter ! I will close this PR for now and I think as you said it will be easier to create a new PR from HEAD.

PR #168, opened 2018-07-24
Use /usr/bin/env bash for more portable shebang lines

benley:
There are still systems widely in use where bash does not exist at `/bin/bash` (e.g. NixOS), or where `/bin/bash` is an extremely outdated version of bash (MacOS).  Using `/usr/bin/env` to locate bash increases portability.  It's not hermetic, but `/bin/bash` was already not hermetic.


benley:
related: https://github.com/bazelbuild/rules_docker/pull/381

Issue #145, opened 2018-06-20
Dry run option


mariusgrigoriu:
I'm wondering if there's interest (or opposition) to an optional parameter to k8s_object that would set the `--dry-run` flag to kubectl. This flag would be useful to validate your objects early in a long-running pipeline, so you don't have to wait until all your infrastructure is up only to find out you have malformed YAML or JSON.

Globegitter:
Yeah it would be nice to have the option for a dry-run, but I wonder if that would be nicer to just be able to pass on at the end of `bazel run`. So `bazel run :target.apply -- --dry-run`, otherwise one would have to modify the `BUILD` file all the time, or have one target for a dry-run and another one for the real apply.

chrislovecnm:
I would rather not modify a BUILD file to get this to run ...

mariusgrigoriu:
Definitely not expecting to modify BUILD file every time you want to turn dry runs on/off. The original idea was to connect the arg to a make variable, but I do like @Globegitter's idea.

mariusgrigoriu:
Oh and there's a new dry-run feature coming in kubectl

chrislovecnm:
This PR will allow us to pass in arbitrary commands https://github.com/bazelbuild/rules_k8s/pull/174

Would that help us out??

Globegitter:
Yep #174 could indeed fix the issue - I left a comment: https://github.com/bazelbuild/rules_k8s/pull/174#discussion_r228050717 as the `$@` in the `apply.sh.tpl ` would have to be moved after the `apply` keyword. But then it does indeed work very nicely.

chrislovecnm:
Can we override arg in a build file rather than a $@?

Globegitter:
@chrislovecnm I just mentioned it in #209 if we name the attribute `arg` then we can set it both in the build file (and bazel will apply the flags at runtime) and via the cli at the end as I showed above.

chrislovecnm:
Closing, as we have this in master now

PR #174, opened 2018-07-31
Add arbitrary commands

omadawn:
Implements #146 

Adds a new variable which will be passed to kubectl as a single string allowing for the insertion of arbitrary arguments not specifically supported by rules_k8s.

Also updated .gitignore to ignore intellij files.


omadawn:
I signed it!

omadawn:
I signed it!

omadawn:
I signed it!

omadawn:
I signed it!

k8s/with-defaults.bzl
```diff
@@ -96,6 +100,7 @@ k8s_defaults = repository_rule(
         "context": attr.string(mandatory = False),
         "user": attr.string(mandatory = False),
         "namespace": attr.string(mandatory = False),
+        "kubectl_args": attr.string(mandatory = False),
```
fejta: I do not think `attr.string` is the best type for this argument. how about `string_list`?

https://docs.bazel.build/versions/master/skylark/lib/attr.html#string_list

k8s/apply.sh.tpl
```diff
@@ -24,4 +24,4 @@ function guess_runfiles() {
 RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
 
 PYTHON_RUNFILES=${RUNFILES} %{resolve_script} | \
-  kubectl --cluster="%{cluster}" --context="%{context}" --user="%{user}" %{namespace_arg} apply -f -
\ No newline at end of file
+  kubectl --cluster="%{cluster}" --context="%{context}" --user="%{user}" %{namespace_arg} %{kubectl_args} apply -f -
```
fejta: We should probably escape `kubectl_args` and all the other inputs. Here's how rules_go does it: https://github.com/bazelbuild/rules_go/blob/368049409d12d0827dcfd0bd401188c22651cc29/go/private/skylib/lib/shell.bzl

Example of loading it: https://github.com/bazelbuild/rules_go/blob/238329474a90d8ac8c128cec3184c515ef467c04/go/private/common.bzl#L18

k8s/with-defaults.bzl
omadawn: Originally I thought this might do best in a hash/associative array (what do they call these in python?) but this is going to contain any number of flags which may or may not contain arguments. I.E.
```--chunk-size=100 --all-namespaces --show-kind=true```

In fact many of the flags themselves may or may not have arguments. all-namespaces can be passed as ```--all-namespaces=true``` or simply ```--all-namespaces```.

So this could go either way. Is there an attr.map or something?

That might give us even better control over the input in regards to escaping strings mentioned below. At some point the keys could be restricted to a list of the allowed arguments in kubectl (some automation to generate said list would obviously be required.)

File: k8s/apply.sh.tpl (hunk 2 above)
Comment (omadawn) [reply k8s/apply.sh.tpl]: I concur on escaping. I'll look at these two examples.

File: k8s/with-defaults.bzl (hunk 1 above)
Comment (jayconrod) [reply k8s/with-defaults.bzl]: There is `attr.string_dict`, but my vote would still be for `string_list`.

My rule-of thumb:

* If the rule needs to do something clever with a particular flag, make it a regular attribute.
* If the flag is just passed to the underlying tool, try to do as little processing as possible and let the underlying tool validate it. `args` is pretty good for this.

Comment (omadawn) [reply k8s/with-defaults.bzl]: Looking through the code and reading a bit on unit testing in Python I don't see how to write unit tests for this. In no small part because almost none of the code I've added is in a standard python file but in .bzl files which I assume get transformed into actual python at some point? 

I see a couple of e2e-test.sh files, mostly under the examples directory. Are you thinking e2e?

My code is just a copy/paste and slight modification of existing code. If there were unit tests of the existing I would have done the same cribbing off the test. ^_^ Does anyone have examples I can look at?

If you're thinking the e2e stuff I've looked at those and I think I can muddle through copying one of the two hello examples.

Comment (nlopezgi) [reply k8s/with-defaults.bzl]: Thanks for taking the time to work on the tests. Unit testing skylark rules is not possible, so yes, we are thinking e2e tests. You should just be able to copy some of the examples (as you say) and add some kubectl_args and verify they behave as expected.

**Comment by omadawn** [comment]
O.K. Added example/hellohttp2 which uses 
```
args = [
    	"--validate=true",
    	"--record=true",
    ],
```
in two of the k8s_deploy targets.

Implemented args which is now passed to the compiled target script.

The functionality worked when I updated WORKSPACE to point to my cluster and namespace and ran say ```bazel run examples/hellohttp2/staging.create``` or ```bazel run examples/hellohttp2/staging.apply```.

Which has exposed a new wrinkle.

Different arguments are valid for different kubectl commands.

So the example provided will fail if you run ```bazel run examples/hellohttp2/staging.delete``` (Neither --validate nor --record are valid for the delete command.

**Comment by omadawn** [comment]
Question, can I determine how this was called within my build? I mean in my BUILD file have something that detects if we were passed ```staging.create``` or ```staging.delete```

If so I could update the example to set @args to a different list of values depending upon the target.

Something like:
```
if target =~ ".create"
@kcargs = [
        "--validate=true",
        "--record=true",
    ],
else
endif

k8s_deploy(
    name = "staging",
    template = "deployment.yaml",
    args = @kcargs,
)
```

Except, you know, actually in bazel's syntax.

**Comment by mariusgrigoriu** [comment]
My take on this is that any args passed to k8s_object needs to work for all kubectl commands, namely the args listed in `kubectl options`. This would match the behavior with other rule args such as `cluster`, `context`, `user`, `namespace` and so on.

**Comment by omadawn** [comment]
Updated the demo and the directions to only include kubectl flags which are global to all kubectl commands.

**Comment by omadawn** [comment]
Is there a way to get more detail about why the test failed? Like console output or something? It just says "exited with error code 1"

**Comment by nlopezgi** [comment]
You should be able to click on the details button next to one of the bazel test (xx platform) checks and then go to the failed test -> target log tab. If you don't see a link or cannot see results after following link let me know to fix. In any case, to unblock you, the info that I see is:
Build log:
...
//examples/hellogrpc:deployment_test                                     PASSED in 0.2s
//examples/hellohttp:deployment_test                                     PASSED in 0.2s
//k8s:resolver_test                                                      PASSED in 28.9s
//examples/hellohttp2:deployment_test                                    FAILED in 3 out of 3 in 0.2s
  Stats over 3 runs: max = 0.2s, min = 0.1s, avg = 0.2s, dev = 0.0s
  /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ce54144dabde77b7400d357492c49be3/execroot/io_bazel_rules_k8s/bazel-out/k8-fastbuild/testlogs/examples/hellohttp2/deployment_test/test_attempts/attempt_1.log
  /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ce54144dabde77b7400d357492c49be3/execroot/io_bazel_rules_k8s/bazel-out/k8-fastbuild/testlogs/examples/hellohttp2/deployment_test/test_attempts/attempt_2.log
  /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ce54144dabde77b7400d357492c49be3/execroot/io_bazel_rules_k8s/bazel-out/k8-fastbuild/testlogs/examples/hellohttp2/deployment_test/test.log
Executed 4 out of 4 tests: 3 tests pass and 1 fails locally.


Test log for //examples/hellohttp2:deployment_test:
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //examples/hellohttp2:deployment_test
-----------------------------------------------------------------------------
E
======================================================================
ERROR: test_things_match (__main__.DeploymentTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ce54144dabde77b7400d357492c49be3/sandbox/linux-sandbox/4/execroot/io_bazel_rules_k8s/bazel-out/k8-fastbuild/bin/examples/hellohttp2/deployment_test.runfiles/io_bazel_rules_k8s/examples/hellohttp2/deployment_test.py", line 26, in test_things_match
    with open(TestData('examples/hellohttp/deployment.yaml'), 'r') as f:
IOError: [Errno 2] No such file or directory: '/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ce54144dabde77b7400d357492c49be3/sandbox/linux-sandbox/4/execroot/io_bazel_rules_k8s/bazel-out/k8-fastbuild/bin/examples/hellohttp2/deployment_test.runfiles/io_bazel_rules_k8s/examples/hellohttp/deployment.yaml'

----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (errors=1)

**Comment by omadawn** [comment]
@nlopezgi Thanks, I'll look at that again once the tests on my latest updates show up.
I realized a lot of the files still pointed to examples/hellohttp instead of hellohttp2 and fixed that.

**Comment by nlopezgi** [comment]
Kicked off presubmits again. Let me know if you want me to kick them off later after more commits.

**Comment by omadawn** [comment]
Thanks, that one was broken to. Fixed and pushed again, hopefully that's the last one and this test goes through. ^_^

**Comment by nlopezgi** [comment]
@fejta could you please help give this PR one more look and provide any more feedback if needed (I've not reviewed it in full and might not be able to get to do a detailed review this week)

**Comment by nlopezgi** [comment]
looks like the travis CI build failed due to this being pinned to an old version of rules_docker that is pointing to a java image that for some reason was deleted (seems fishy the distroless images got deleted, will try to find out why later). I'll send a PR to update pin to rules_docker and update your branch once that is in.

**Comment by omadawn** [comment]
Thank you. I was baffled.

**Comment by nlopezgi** [comment]
@omadawn  Sorry for the long delay to get back. Had to fix some issues with CI that were resulting in spurious errors. Just ran the tests again and they seem to legitimately fail:
Error from server (NotFound): namespaces "metalstorm" not found
Please fix and let me know when its ready to kick off tests again.

**Comment by omadawn** [comment]
Thank you. Stupid mistake. I changed a variable so that I could run the test locally and then forgot to set namespace back to $USER so that it runs in the pipeline.  I remembered to update the other places where that happens. ^_^

Fixed

**Comment by omadawn** [comment]
Looks like some other change has been merged into master, re-based.

**Comment by nlopezgi** [comment]
this is due to a change in the bazel 0.17.0 rc1 release which impacts this repo due to its dependency on rules_docker. Have already fixed upstream (https://github.com/bazelbuild/rules_docker/pull/501), will send a PR to update pin to rules_docker, I'll merge your branch and retest again, sorry for this issue.

**Comment by omadawn** [comment]
Sweet, thanks.

**Comment by nlopezgi** [comment]
you can update branch and re-test once https://github.com/bazelbuild/rules_k8s/pull/180 is in

**Comment by omadawn** [comment]
Thanks

**Comment by fejta** [comment]
Is this PR alive?

**Comment by omadawn** [comment]
Still alive but keeps failing due to things outside of my control so I haven't been following up on it as much as I probably should have.

**Comment by fejta** [comment]
The errors appear to be in your PR?

```console
bazel-bin/examples/hellohttp2/py/staging.apply
...
Checking response from service: DEMO  matches: DEMO30997<space>
```

You are changing these files

chrislovecnm:
Fixes: https://github.com/bazelbuild/rules_k8s/issues/190

nlopezgi:
hi @omadawn , have you had a chance to look at the test issues? thanks!

chrislovecnm:
@nlopezgi let us know if someone can assist with this PR this featuure would help a TON with these rules.

Globegitter:
yep @omadawn the problem is `$RESOURCE_NAME`, `RESOURCE_NAME='deployment.apps/hello-grpc-staging created (dry run)'` so the command it is trying to execute is `kubectl --cluster=gke_rules-k8s_us-central1-f_testing --context=gke_rules-k8s_us-central1-f_testing --user= --namespace=markus describe deployment 'deployment.apps/hello-grpc-staging created (dry run)'`

Do you think you have time to fix the issues in the next couple of days? Otherwise I am happy to look into this and finish it off.

k8s/apply.sh.tpl
```diff
@@ -24,4 +24,4 @@ function guess_runfiles() {
 RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
 
 PYTHON_RUNFILES=${RUNFILES} %{resolve_script} | \
-  kubectl --cluster="%{cluster}" --context="%{context}" --user="%{user}" %{namespace_arg} apply -f -
\ No newline at end of file
+  kubectl --cluster="%{cluster}" --context="%{context}" --user="%{user}" %{namespace_arg} $@ apply -f -
```
Globegitter: @omadawn Any reason you put this `$@` before the `apply`? I was trying to run `bazel-bin/examples/hellohttp2/py/staging.apply --dry-run` and the way this currently is setup this does not work. If the `$@` is moved after the `apply` then the dry-run flag does indeed work. Would be nice to get #145 fixed.

omadawn: That won't actually work.
The problem is that flags like --dry-run are not global. It works for apply but not for say describe. With the current architecture of the plugin all flags are applied to all options. We would need to provide a way to specify individual settings for different activities.

chrislovecnm:
@omandan I actually used your code and submitted a different PR. Not sure how you want to proceed

omadawn:
> @omandan I actually used your code and submitted a different PR. Not sure how you want to proceed

I saw that, I'm happy to let you proceed. I was just in responding to a comment and I figured I'd hit retest Since I hadn't after the build pipeline got fixed.

chrislovecnm:
@omadawn I would very much appreciate a code review.

fejta:
Safe to clone this PR then in favor of #198?

omadawn:
> Safe to clone this PR then in favor of #198?

I don't know. It works, I'm using it in prod. But my tests are failing and I have no idea how to figure out why. The output of the test runner just says it failed.
```
FAIL: //examples/hellohttp2:deployment_test (see /root/.cache/bazel/_bazel_prow/0996f4cdf646bf91686374917a22c2ea/execroot/io_bazel_rules_k8s/bazel-out/k8-fastbuild/testlogs/examples/hellohttp2/deployment_test/test.log)
```
I would be happy to dig in and see why but I don't know how to access that. It's not available from gubernator.

Chris's change has been more active lately and I'm happy to see someone else go through tis frustration instead of me. One of them should eventually make it through.

k8s/apply.sh.tpl
Globegitter: But do the global flags not also work behind the `apply/describe` command? Looking at the test `kubectl apply/describe --v=0 --alsologtostderr=true --stderrthreshold=1` seems to work for me also.

Globegitter:
If we do need two separate ways of specify flags, once global flags and once command specific flags I thiink it would make sense two split this out into two, `kubectl_args` could be global ones and `args` command specific, so then one can also use the `bazel run :target.apply -- --dry-run` way of calling. WDYT?

Globegitter:
@omadawn I am happy to take over this PR to try and finish it off - I opened up #207 for that. Let me know if that is ok with you - happy to also close it again :)

omadawn:
> There is a difference though between these two PRs `args` is actually a bazel specific attribute, so that bazel then applies the arguments on run, where-as `kubectl_args` is not. So I do think if we get something to work with `args` that would probably fit better, no?

I switched from kubectl_args to args because args was a more common pattern. But if it's intended to apply to command line arguments to bazel I don't fully understand how you pass those when running bazel.

I would say the 'best' fit would be to follow a built in bazel convention unless I'm using some common bazel thing the wrong way.

omadawn:
> If we do need two separate ways of specify flags, once global flags and once command specific flags I thiink it would make sense two split this out into two, `kubectl_args` could be global ones and `args` command specific, so then one can also use the `bazel run :target.apply -- --dry-run` way of calling. WDYT?

If you can pass things that way then my vote is use args and move it to the right of the kubectl command. 

```kubectl <stuff> create $@ --```

I have previously argued against moving this to the right of the command to support flags like --dry-run because these flags are kubectl command specific. I.E. --dry-run is not applicable for describe and generates an error. But that was when I thought you could only use this within the build file and would have to modify it to add or remove this flag.

What about both?

```kubectl %(kubectl_global_args) create $@ --```
 
The former would be only for args which are global to all commands and thus intended to be used in the build file. 

The latter would provide the command line arguments passed to bazel and thus be command specific.

nlopezgi:
Closing this with https://github.com/bazelbuild/rules_k8s/pull/224 proposed as a solution to the required use case. Please comment in https://github.com/bazelbuild/rules_k8s/issues/209 if you still feel we need to keep this open for some reason.

PR #176, opened 2018-08-15
update rules_docker pin

nlopezgi:


nlopezgi:
@fejta Prow seems to be behaving strangely. Its failed several times getting external resources, its taking way too long to run, and I cant see the logs as it runs when I follow the details link. Do you know if anything  is up?

nlopezgi:
Trying to open a new PR to see if that helps with prow issues, this one is reporting errors when there is a build still running, links are changing from broken locations after some time. ugh.

PR #179, opened 2018-08-20
 update rules_docker pin 

nlopezgi:
Lets try to send this PR again


nlopezgi:
/test pull-rules-k8s-e2e
Upsized our testing cluster as it seems presubmits were OOM

PR #184, opened 2018-08-26
Added 'namespace_file' attribute

excavador:
I added ability to specify 'namespace_file' in addition to namespace.

### DIFF
--- a/k8s/object.bzl
+++ b/k8s/object.bzl
@@ -167,6 +167,11 @@ def _common_impl(ctx):
     namespace_arg = "$(cat %s)" % _runfiles(ctx, namespace_file)
     files += [namespace_file]
 
+  if ctx.file.namespace_file:
+    namespace_file = ctx.file.namespace_file
+    namespace_arg = "$(cat %s)" % _runfiles(ctx, namespace_file)
+    files += [namespace_file]
+
   if namespace_arg:
     namespace_arg = "--namespace=\"" +  namespace_arg + "\""
 
@@ -201,6 +206,7 @@ def _common_impl(ctx):
 
 _common_attrs = {
     "namespace": attr.string(),
+    "namespace_file": attr.label(allow_single_file=True),
     # We allow cluster to be omitted, and we just
     # don't expose the extra actions.
     "cluster": attr.string(),
@@ -424,6 +430,12 @@ def k8s_object(name, **kwargs):
   _reversed(name=name + ".reversed", template=kwargs.get("template"),
             visibility=kwargs.get("visibility"))
 
+  namespace = kwargs.pop("namespace", None)
+  namespace_file = kwargs.pop("namespace_file", None)
+  if namespace and namespace_file:
+    fail("you should choose one: 'namespace' or 'namespace_file'")
+
+
   if "cluster" in kwargs or "context" in kwargs:
     _k8s_object_create(
         name=name + ".create",
@@ -432,7 +444,8 @@ def k8s_object(name, **kwargs):
         cluster=kwargs.get("cluster"),
         context=kwargs.get("context"),
         user=kwargs.get("user"),
-        namespace=kwargs.get("namespace"),
+        namespace=namespace,
+        namespace_file=namespace_file,
         visibility=kwargs.get("visibility"),
     )
     _k8s_object_delete(
@@ -442,7 +455,8 @@ def k8s_object(name, **kwargs):
         cluster=kwargs.get("cluster"),
         context=kwargs.get("context"),
         user=kwargs.get("user"),
-        namespace=kwargs.get("namespace"),
+        namespace=namespace,
+        namespace_file=namespace_file,
         visibility=kwargs.get("visibility"),
     )
     _k8s_object_replace(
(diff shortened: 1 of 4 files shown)


k8s/object.bzl
```diff
@@ -133,6 +133,9 @@ def _resolve(ctx, string, output):
   )
 
 def _common_impl(ctx):
+  if ctx.attr.namespace and ctx.attr.namespace_file:
```
nlopezgi: Validation / failure should be done in macro: https://github.com/bazelbuild/rules_k8s/blob/master/k8s/object.bzl#L416

k8s/object.bzl
```diff
@@ -159,11 +162,18 @@ def _common_impl(ctx):
...
+  namespace_file = None
+
   namespace_arg = ctx.attr.namespace
   namespace_arg = ctx.expand_make_variables("namespace", namespace_arg, {})
   if "{" in ctx.attr.namespace:
     namespace_file = ctx.new_file(ctx.label.name + ".namespace-name")
     _resolve(ctx, ctx.attr.namespace, namespace_file)
+
+  if ctx.file.namespace_file:
+    namespace_file = ctx.file.namespace_file
```
nlopezgi: If you dont add the namespace_file to the files, will it still be available in the struct returned below in line 210? If not, then this will result in failed builds for users that run with sandboxing enabled

fejta:
> Could you explain a bit more why you need to add this option to specify namespace from a file?

👍 any insights here?

excavador:
The reason are simple: I am generating namespace name based on git branch name:
```
def _impl(ctx):
    output = ctx.actions.declare_file(ctx.label.name)
    ctx.actions.run_shell(
        inputs = [ctx.file.head],
        outputs = [output],
        command = """#!/bin/bash
set -euo pipefail
cd $(dirname $(dirname $(realpath {head})))
git rev-parse --abbrev-ref HEAD > {output}
""".format(head = ctx.file.head.path, output = output.path),
    )
    return [DefaultInfo(files = depset([output]))]

git_branch_name = rule(
    implementation = _impl,
    attrs = {
        "head": attr.label(allow_single_file = True, default = "@self//:.git/HEAD"),
    },
)

...
git_branch_name(
    name = "gbn",
)
...
k8s_object(
   name ="some_name"
   namespace_file = ":gbn"
```

k8s/object.bzl
excavador: > If you dont add the namespace_file to the files, will it still be available in the struct returned below in line 210? If not, then this will result in failed builds for users that run with sandboxing enabled

This file also generated above
```
  if "{" in ctx.attr.namespace:
    namespace_file = ctx.new_file(ctx.label.name + ".namespace-name")
    _resolve(ctx, ctx.attr.namespace, namespace_file)
```

I just extended the original behavior: upstream version allows to specify namespace name as constant or use stamped variables. Now you can specify name based on file content which generated by bazel

k8s/object.bzl
excavador: Will fix, thank you!

k8s/object.bzl
excavador: Fixed

k8s/object.bzl
excavador: Fixed

excavador:
I added example

fejta:
a) I am not sure this is the best way to address the need
b) I'm not really sure this is something we need upstream

If you look at something like how we create image tags in kubernetes/test-infra: https://github.com/kubernetes/test-infra/blob/24a930063493c4af1af7f35838c7a780a07b774a/image.bzl#L36

you'll see we use variables like `latest-{BUILD_USER}` that are expanded:
```python
outs['%s:{DOCKER_TAG}' % img] = target
outs['%s:latest-{BUILD_USER}' % img] = target
```

We have a script that produces these values (for ones which aren't auto-generated by bazel):
https://github.com/kubernetes/test-infra/blob/master/hack/print-workspace-status.sh#L20-L30

We configured our `.bazelrc` file to automatically run this workspace command:
https://github.com/kubernetes/test-infra/blob/24a930063493c4af1af7f35838c7a780a07b774a/.bazelrc#L1

And we already support this kind of expansion for the namespace value:
https://github.com/bazelbuild/rules_k8s/blob/master/k8s/object.bzl#L162-L168

So I do not think this PR is necessary

excavador:
@fejta it's up to you. 

We need it for Continuous Delivery, and for setup namespace like "<aws-identity-name>-<git_branch_name>"
For simple cases like you described enough workspace state command.

For more complex cases I need to manipulation in BUILD rules, like execution my golang binaries.

excavador:
So, please reject PR if it's unnecessary :)
If you want, we can discuss this case on Bazel Conference

fejta:
Can you give an example of something you cannot do with templating?


`k8s_object(namespace='{AWS_IDENTITY}-{GIT_BRANCH}')`

appears to do exactly what you want

the values of `AWS_IDENTITY` and `GIT_BRANCH` are calculated each time bazel runs. So it is not obvious to me what functionality a file adds that isn't already covered here.

excavador:
@fejta for calculate AWS_IDENTITY I need to build golang binary from my repository and execute it.
I can not build any binary BEFORE execute workspace status command

nlopezgi:
@excavador thanks for your clarification. Would https://github.com/bazelbuild/rules_k8s/issues/190 also solve your use case?

excavador:
@nlopezgi as workaround - yes :)

PR #187, opened 2018-10-02
Allow project override

fejta:
Allows `PROJECT=my-project make push` in case someone wants to push this image to their own project


fejta:
Thanks for the quick review!

PR #196, opened 2018-10-18
Use kubectl describe foo name

fejta:
If we do not include the `-o name` then the command output includes a ` (dry run)` suffix.

/assign @smukherj1 @nlopezgi 

This seems to be causing issues in recent kubectl versions (https://github.com/bazelbuild/rules_k8s/pull/193#issuecomment-431026365).

* Formatting fixes while I'm in these files
* Run `delete || true` before running create, in case zombie resources exist from a prior run.
* Logging the steps before they run to assist debugging

### DIFF
--- a/examples/hellogrpc/e2e-test.sh
+++ b/examples/hellogrpc/e2e-test.sh
@@ -24,44 +24,49 @@ if [[ -z "${1:-}" ]]; then
 fi
 
 get_lb_ip() {
-    kubectl --namespace=${USER} get service hello-grpc-staging \
-	-o jsonpath='{.status.loadBalancer.ingress[0].ip}'
+  kubectl --namespace=${USER} get service hello-grpc-staging \
+    -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
 }
 
 # Ensure there is an ip address for hell-grpc-staging:50051
 apply-lb() {
+  echo Applying service...
   bazel build examples/hellogrpc:staging-service.apply
   bazel-bin/examples/hellogrpc/staging-service.apply
 }
 
 create() {
-   bazel build examples/hellogrpc/${LANGUAGE}/server:staging.create
-   bazel-bin/examples/hellogrpc/${LANGUAGE}/server/staging.create
+  echo Creating $LANGUAGE...
+  bazel build examples/hellogrpc/${LANGUAGE}/server:staging.create
+  bazel-bin/examples/hellogrpc/${LANGUAGE}/server/staging.create
 }
 
 check_msg() {
-   bazel build examples/hellogrpc/${LANGUAGE}/client
+  bazel build examples/hellogrpc/${LANGUAGE}/client
 
-   OUTPUT=$(./bazel-bin/examples/hellogrpc/${LANGUAGE}/client/client $(get_lb_ip))
-   echo Checking response from service: "${OUTPUT}" matches: "DEMO$1<space>"
-   echo "${OUTPUT}" | grep "DEMO$1[ ]"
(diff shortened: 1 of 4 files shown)


erain:
I thought `kubectl get foo name` is more commonly used than `kubectl get foo/name`?  Example: [kubectl Cheat Sheet](https://kubernetes.io/docs/reference/kubectl/cheatsheet/)

fejta:
Maybe `kubectl get foo name` and `kubectl get foo/name` are both fine, but it doesn't like `kubectl get foo foo/name`

fejta:
Looks like the problem is in the actual bazel rules... Let me see if I can figure this out...

fejta:
These e2e tests are the WORST :sweat_smile:

fejta:
Okay I think I have it!

Looks like the problem here is that `kubectl create --dry-run -f "%{unresolved}"` now outputs (dry run):
`deployment.apps/hello-grpc-staging created (dry run)`
But we want the name without the (dry run) suffix:
`deployment.apps/hello-grpc-staging created`

We can fix this by adding an `-o name` flag.

fejta:
PASS! :fireworks: :champagne: :confetti_ball:

examples/hellohttp/e2e-test.sh
```diff
@@ -40,33 +41,35 @@ check_msg() {
...
-    else
-	echo "Bad substitution should fail!"
-	exit 1
-    fi
+  echo Checking a bad substitution...
+  if ! bazel run examples/hellohttp:error-on-run; then
+    echo "Success, substitution failed."
+    return 0
+  else
+    echo "Bad substitution should fail!"
```
chrislovecnm: If I am a grammar critic it could read a tad better, but a nit pick

examples/hellohttp/e2e-test.sh
```diff
@@ -40,33 +41,35 @@ check_msg() {
 }
 
 edit() {
-   ./examples/hellohttp/${LANGUAGE}/edit.sh "$1"
+  echo Setting $LANGUAGE to "$1"
```
chrislovecnm: Please quote language var. We might want to lint with shellcheck

examples/hellogrpc/e2e-test.sh
```diff
@@ -24,44 +24,49 @@ if [[ -z "${1:-}" ]]; then
 fi
 
 get_lb_ip() {
-    kubectl --namespace=${USER} get service hello-grpc-staging \
-	-o jsonpath='{.status.loadBalancer.ingress[0].ip}'
+  kubectl --namespace=${USER} get service hello-grpc-staging \
```
chrislovecnm: Please put quotes around all variables

examples/todocontroller/e2e-test.sh
```diff
@@ -18,66 +18,72 @@ set -e
 LANGUAGE="$1"
 
 function get_lb_ip() {
-    kubectl --namespace=${USER} get service hello-grpc-staging \
-	-o jsonpath='{.status.loadBalancer.ingress[0].ip}'
+  kubectl --namespace=${USER} get service hello-grpc-staging \
+    -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
```
chrislovecnm: Again missing quotes around a bash var

chrislovecnm:
Well that did not work 😂😂

chrislovecnm:
I can do a follow-up PR as well. @fjeta thoughts?

chrislovecnm:
@fejta I removed my blocking review, so up to you if you want to fix my recommendations or not.  I can do a follow-up PR as I mentioned.

chrislovecnm:
This is blocking testing big time. @erain / @fejta / @nlopezgi can someone merge?

PR #207, opened 2018-10-26
Add command specific arguments

### LINKED ISSUE #145: Dry run option

Globegitter:
This is instead of #174 to be able to pass command specific arguments directly from the cli, e.g. `bazel run :target.apply -- --dry-run`. This would fix #145 and could complement ##198 to be able to specify both kubectl global and command specific flags. Hopefully this will also fix the tests.

Let me know what you think about this approach.


fejta:
> all these commits are going to cause a mess in master

Won't clicking "squash and merge" make github put them into a single commit? Not sure if we need/want to follow the kubernetes org strategy of allowing PRs with multiple commits (and therefore requiring manual commit management).

> Please rebase

The way to do this is to do something like `git rebase -i HEAD~31` and then change all the `pick` lines except the first one to `f` (for `fixup`, which drops the commit message) or `s` (for `squash`, which adds the commit message).

Alternatively you could do `git reset HEAD~31 && git add -A && git commit -m "blah blah blah"`

chrislovecnm:
@fejta this commit will not squash or merge cleanly in a GH submit. I tried to rebase it, not fun.

nlopezgi:
Closing this with https://github.com/bazelbuild/rules_k8s/pull/224 proposed as a solution to the required use case. Please comment in https://github.com/bazelbuild/rules_k8s/issues/209 if you still feel we need to keep this open for some reason.

PR #210, opened 2018-10-31
Kubectl via toolchains

alex1545:
Pass value of toolchain attribute that is the default kubectl path to templates.


nlopezgi:
@chrislovecnm testing of toolchains (other than just verifying rules are not broken) is not possible. Documentation will come later once we have a full solution for toolchains in place (i.e., this PR is work in progress towards getting a full k8s toolchain, we decided to split up the work in small PRs to make review simpler)

PR #214, opened 2018-11-06
Fix getting the current cluster name in README.md

### LINKED ISSUE #213: Unable to deploy to local cluster
I have Kubernetes setup with Docker for Mac.
Bazel k8s is configured:

```
k8s_defaults(
  name = "k8s_deploy",
  kind = "deployment",
  cluster = "docker-for-desktop",
  namespace = "{BUILD_USER}",
)

k8s_deploy(
  name = "user_data_deployment",
  template = ":deployment.yaml",
  images = {
    "user_data:dev": ":image"
  },
)
```

Cluster is up and running locally:

```
$ kubectl config current-context
docker-for-desktop

$ kubectl cluster-info
Kubernetes master is running at https://localhost:6443
KubeDNS is running at https://localhost:6443/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
```

However, I get an error when running the deployment:

```
$ bazel run //backend/deploy/k8s:user_data_deployment.apply
INFO: Analysed target //deploy/k8s:user_data_deployment.apply (0 packages loaded).
INFO: Found 1 target...
Target //deploy/k8s:user_data_deployment.apply up-to-date:
  bazel-bin/deploy/k8s/user_data_deployment.apply
INFO: Elapsed time: 0.476s, Critical Path: 0.01s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
error: cluster "docker-for-desktop" does not exist
close failed in file object destructor:
sys.excepthook is missing
lost sys.stderr
```

Is there any way to use these rules with a locally-running k8s deployment?

pierreis:
The current documentation suggests using the current context name (provided by `kubectl config current-context`) as cluster name.

This does not work in configurations in which there is a mismatch between the cluster and context name (_e.g., K8s provided by Docker for Mac).

This pull request fixes this by using the output of `kubectl config view --minify -o=jsonpath='{.contexts[0].context.cluster}'` instead.

Fixes #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ k8s_defaults(
   name = "k8s_deploy",
   kind = "deployment",
   # This is the name of the cluster as it appears in:
-  #   kubectl config current-context
+  #   kubectl config view --minify -o=jsonpath='{.contexts[0].context.cluster}'
   cluster = "my-gke-cluster",
 )
 ```
@@ -582,7 +582,7 @@ A repository rule that allows users to alias `k8s_object` with default values.
         <p>The name of the cluster to which <code>create, replace, delete,
            describe</code> should speak.</p>
 	<p>This should match the cluster name as it would appear in
-           <code>kubectl config current-context</code></p>
+           <code>kubectl config view --minify -o=jsonpath='{.contexts[0].context.cluster}'</code></p>
       </td>
     </tr>
     <tr>


README.md
```diff
@@ -106,7 +106,7 @@ k8s_defaults(
   name = "k8s_deploy",
   kind = "deployment",
   # This is the name of the cluster as it appears in:
-  #   kubectl config current-context
+  #   kubectl config view --minify -o=jsonpath='{.contexts[0].context.cluster}'
```
fejta: @nlopezgi 
Context is the combination of user + cluster. By default gcloud will set user, cluster and context to be the same thing (probably why this bug exists) but that need not be the case. Totally possible that other distros besides GKE do something different.

Context may actually be what most people want here (and is typically what people use). Probably worth providing a `context` parameter here in addition to `cluster` (which is probably better than repurposing cluster to be cluster).

@mattmoor can you comment on the history here? Is this an accident or intentional?

fejta:
/assign @smukherj1 

Any idea why this keeps failing? Are we out of quota?

fejta:
Also @smukherj1 @nlopezgi can you trigger the non prow builds?

smukherj1:
Opened #216 to hopefully reduce e2e flakiness

PR #219, opened 2018-11-08
Disable flakiest part of e2e test

smukherj1:


chrislovecnm:
> What could go wrong

@fejta you had to go and say that

chrislovecnm:
Is your echo causing the failure?

smukherj1:
@chrislovecnm No a completely different bit is failing. It's this line that is printing a different output than expected: https://github.com/bazelbuild/rules_k8s/blob/master/examples/todocontroller/e2e-test.sh#L50

chrislovecnm:
> @chrislovecnm No a completely different bit is failing. It's this line that is printing a different output than expected: https://github.com/bazelbuild/rules_k8s/blob/master/examples/todocontroller/e2e-test.sh#L50

🤦‍♂️

smukherj1:
That was in the bazel test, not the end to end tests. I feel like we should look into splitting up the tests so that they can be rerun in smaller chunks. The failure probability across the monolithic chunk is too high

smukherj1:
`
kubectl create namespace build-400
The connection to the server 35.184.155.32 was refused - did you specify the right host or port?
`

huh? Gonna check the quotas real quick...

smukherj1:
Quotas look fine.. this is infuriating...

smukherj1:
-____-

pierreis:
The line that failed is now https://github.com/bazelbuild/rules_k8s/blob/d2ad622f18a07d367dfc287cb7635132896c6bff/examples/todocontroller/e2e-test.sh#L80

chrislovecnm:
@smukherj1 don't stop now, you are almost there 😄

chrislovecnm:
@smukherj1 / @fejta any idea why the deletes are failing??

smukherj1:
@chrislovecnm The delete didn't fail here. It's this line https://github.com/bazelbuild/rules_k8s/blob/a63807a37b3e42708b4b3892f493cea9eb3da2f3/examples/todocontroller/e2e-test.sh#L52 that failed which triggered a call to delete because we set a trap to delete here https://github.com/bazelbuild/rules_k8s/blob/a63807a37b3e42708b4b3892f493cea9eb3da2f3/examples/todocontroller/e2e-test.sh#L109

The delete executed fine. Need to figure out why the expected output isn't appearing in the describe command

chrislovecnm:
@smukherj1 really appreciate you fixing this!!!!!!!  

I need to add more e2e tests, and I am wondering if I should create an entire new test that runs by itself. 🤔

smukherj1:
Hmm my logic to detect whether the controller update worked didn't behave as expected. Will continue looking at this later

pierreis:
Informatively, what is the added value of the todo controller test compared to the http and grpc ones? Is it conceivable to disable it until we can figure out a proper solution?

chrislovecnm:
@pierreis agreed you want to do a PR?

chrislovecnm:
@pierreis already working on a PR

pierreis:
Awesome, thanks

chrislovecnm:
@pierreis https://github.com/bazelbuild/rules_k8s/pull/222

examples/todocontroller/e2e-test.sh
```diff
@@ -47,7 +47,24 @@ function create() {
...
+    echo "${OUTPUT}" | grep "Done:[ ]*true"
+    set -o errexit
+    retval=$?
+    if [ $retval -ne 0 ]; then
+      echo "Controller update isn't done yet. Sleeping 60s..."
+      sleep 60
+    else
+      break
+    fi
+  done;
```
fejta: semicolon not necessary

examples/todocontroller/e2e-test.sh
```diff
@@ -47,7 +47,24 @@ function create() {
...
+  for iattempt in $(seq 1 $NUM_ATTEMPTS); do
+    OUTPUT=$(./bazel-bin/examples/todocontroller/example-todo.describe)
+    echo "Attempt ${iattempt}/${NUM_ATTEMPTS}: Checking whether server is done with update"
+    # Disable errexit temporarily because we expect the following command to
+    # fail a few times
+    set +o errexit
+    echo "${OUTPUT}" | grep "Done:[ ]*true"
+    set -o errexit
+    retval=$?
+    if [ $retval -ne 0 ]; then
```
fejta: an if statement, in a for loop, in a function... Starting to trigger my "excessively complicated for bash, rewrite in python" sensor :)

examples/todocontroller/e2e-test.sh
```diff
@@ -47,7 +47,24 @@ function create() {
...
-  OUTPUT=$(./bazel-bin/examples/todocontroller/example-todo.describe)
+  NUM_ATTEMPTS=10
+  OUTPUT=""
+  for iattempt in $(seq 1 $NUM_ATTEMPTS); do
+    OUTPUT=$(./bazel-bin/examples/todocontroller/example-todo.describe)
+    echo "Attempt ${iattempt}/${NUM_ATTEMPTS}: Checking whether server is done with update"
+    # Disable errexit temporarily because we expect the following command to
+    # fail a few times
+    set +o errexit
+    echo "${OUTPUT}" | grep "Done:[ ]*true"
```
fejta: Most of this can be simplified to something like: `echo $OUTPUT | grep Done && return 0 || true`

examples/todocontroller/e2e-test.sh
```diff
@@ -47,7 +47,24 @@ function create() {
...
+    set -o errexit
+    retval=$?
```
fejta: Another way would be:
```bash
if ! echo $OUT | grep Done; then 
  echo "Sleeping..."
else
  break
fi
```

chrislovecnm: What we actually need is kubectl rollout 😆

smukherj1:
Abandoning as this test was completely disabled by #222

PR #224, opened 2018-11-10
Arbitrary args for kubectl

nlopezgi:
Trying this PR to replace all other PRs discussed in https://github.com/bazelbuild/rules_k8s/issues/209.
With this PR:
- users can add args (string list) to their k8s_deploy object
- users can use bazel run <target> -- <extra args>
- modified templates so the kubectl command is printed out when running bazel run <target>, useful for validation in e2e tests

Hope this satisfies everyone's needs.
fyi @chrislovecnm @Globegitter @omadawn

PR #225, opened 2018-11-14
Arbitrary arguments for k8s_objects

samschlegel:
This builds on #224 by passing through arguments to the individual runs


nlopezgi:
Hi @samschlegel ,
thanks for sending this PR.
To test this you could add to https://github.com/bazelbuild/rules_k8s/blob/master/examples/todocontroller/e2e-test.sh the same functionality like the one I added in https://github.com/bazelbuild/rules_k8s/pull/224 to https://github.com/bazelbuild/rules_k8s/blob/master/examples/hellohttp/e2e-test.sh to perform checks. Then add a test like https://github.com/bazelbuild/rules_k8s/blob/master/examples/hellohttp/e2e-test.sh#L118 in https://github.com/bazelbuild/rules_k8s/blob/master/examples/todocontroller/e2e-test.sh that runs "bazel run examples/todocontroller: everything -- -v=2" and checks the command(s) executed (should be printed as a result of the changes in https://github.com/bazelbuild/rules_k8s/pull/224 in the tpl files) do(es) contain the -v=2 arg as you expect

chrislovecnm:
@samschlegel to add to what Nick mentioned. Please do not enable the controller tests. We have disabled those till we work some problems out of them. You can do the run everything on the other examples.

samschlegel:
So I should add them to todocontroller, but not enable the tests?

samschlegel:
It looks like hello_grpc also uses k8s_objects, so I could do it there instead.

chrislovecnm:
@samschlegel a grpc test would be great!

samschlegel:
Test added!

chrislovecnm:
Can you rebase?

examples/hellogrpc/e2e-test.sh
```diff
@@ -88,6 +109,18 @@ delete() {
...
 
+# e2e test that checks that args are added to the kubectl apply command
+check_kubectl_args() {
+    # Checks that bazel run <some target> does pick up the args attr and
+    # passes it to the execution of the template
+    EXPECT_CONTAINS "$(bazel run examples/hellogrpc:staging.apply)" "apply --v=2"
+    # Checks that bazel run <some target> -- <some extra arg> does pass both the
+    # args in the attr as well as the <some extra arg> to the execution of the
+    # template
+    EXPECT_CONTAINS "$(bazel run examples/hellogrpc:staging.apply -- --v=1)" "apply --v=2 --v=1"
```
chrislovecnm: Not sure if this is your test, but we have two —v on it

samschlegel: I just copied how @nlopezgi did the e2e test in hellohttp

nlopezgi:
ugh, mac builds are timing out pulling images, will have to figure out tomorrow what's the problem

PR #228, opened 2018-11-20
Include __init__.py files

samschlegel:
We are currently using our own custom resolver that imports `@io_bazel_rules_k8s//k8s:resolver_lib` to run the default resolve pass after our own custom passes. This works fine when our resolver uses `par_binary`, but fails when we use `py_binary`. For whatever reason, it seems that the files created by `py_binary.legacy_create_init` aren't included in the runfiles of the `k8s_object` rules. 

There are three ways this could be fixed:
1. Fix it upstream by including the `__init__.py` files in `py_binary.attrs.runfiles`
2. Change the resolver script to run the resolver executable using the `execpath` Make variable.
3. Create and include `__init__.py` files (This PR)

Since the expected behaviour moving forward is to include these `__init__.py` files [0], I feel like 3's probably the best option for now, however it seems that 2 might be a better solution

[0] https://docs.bazel.build/versions/master/be/python.html#py_binary.legacy_create_init


samschlegel:
Added licenses and rebased!

chrislovecnm:
@nicklopezgi can you get tests running? Is this change ok?

nlopezgi:
overall looks good to me, I dont have any way to test the specific use case, but as long as the e2e tests are passing I think this should be fine. Kicked them off.

Issue #234, opened 2018-11-26
Add documentation for kubectl toolchain changes


samschlegel:
The changes in #229 cause a breakage if `rules_go` has not been pulled in, and this is not currently mentioned in the Getting Started. It is also non obvious as to how to disable building kubectl from source, for those who wish to rely on the installed kubectl version.

@nlopezgi Mentioned there was a doc PR incoming, but figured I'd add this for tracking

nlopezgi:
thanks for creating this, will work on this PR now before I forget.

chrislovecnm:
Should we be building kubectl from source by default?

nlopezgi:
I changed the default for users just adding k8s_repositories() to their WORKSPACE to get the "use the one from path" to stay with the behavior before #229. Please see the PR and let me know if anyone has thoughts about it.

PR #235, opened 2018-11-27
Update toolchain docs

nlopezgi:
* also make default (when importing these rules) to use kubectl from path (simpler atm, less error prone)
* provide instructions for how to use the new option to build kubectl from sources
* also show a nice error message if no kubectl tool is found when user attempts to do `bazel run` for a target that requires the toolchain is configured

https://github.com/bazelbuild/rules_k8s/issues/234

### DIFF
--- a/k8s/k8s.bzl
+++ b/k8s/k8s.bzl
@@ -12,38 +12,46 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Rules for manipulation of K8s constructs."""
+
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load(":with-defaults.bzl", "k8s_defaults")
 load("//toolchains/kubectl:kubectl_configure.bzl", "kubectl_configure")
 
-def k8s_repositories():
-  """Download dependencies of k8s rules."""
+def k8s_repositories(build_kubectl_srcs = False):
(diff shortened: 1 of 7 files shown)


nlopezgi:
@fejta any idea why this is showing me I can squash and merge even though checks have not passed?

k8s/k8s.bzl
```diff
@@ -12,38 +12,46 @@
...
-  # Register the default kubectl toolchain targets for supported platforms
-  # note these work with the autoconfigured toolchain
-  native.register_toolchains(
-    "@io_bazel_rules_k8s//toolchains/kubectl:kubectl_linux_toolchain",
-    "@io_bazel_rules_k8s//toolchains/kubectl:kubectl_osx_toolchain",
-    "@io_bazel_rules_k8s//toolchains/kubectl:kubectl_windows_toolchain",
-  )
+    # Register the default kubectl toolchain targets for supported platforms
+    # note these work with the autoconfigured toolchain
+    native.register_toolchains(
```
smukherj1: Wondering if we should move the toolchain registration to be under the "k8s_config" if condition as well? Allows users to decide their own toolchain registration priority order when defining their own toolchain configuration

nlopezgi: i think it makes sense, will do in an upcoming PR

Issue #245, opened 2018-12-11
document better how to use custom kubernetes version in k8s config


nlopezgi:
- Provide an example in readme docs
- ideally add an e2e test that pulls a custom version (as a subfolder / subproject in this repo)
- fail early if user does not provide k8s_prefix / k8s_sha256

greggdonovan:
An example for kubectl version 1.10 would be very helpful for us.

smukherj1:
@greggdonovan Documenting how to use a kubectl toolchain where kubectl was built from source in #251.

Unfortunately, v1.10.1 can't be built from source because I get the error mentioned in https://github.com/bazelbuild/rules_k8s/pull/251#issuecomment-448381811

I verified building v1.13.1 from source works. We don't support prebuilt binaries yet.

chrislovecnm:
@greggdonvan we might be able to backport into kubernetes the changes that have been made for 1.13 to fix bazel, or make a branch for Etsy.

@smukherj1 what is the min version that works? Do we have that documented?

smukherj1:
We added a defaults.bzl in the toolchains directory and the documentation points to this as the currently recommended version that is known to work. The documentation of the kubectl_configure function mentions v1.13.0-beta.1 is the minimum version known to work. The example in the docs (being added by my PR) shows it also works with v1.13.1. The documentation says it should work with v1.13.0-beta.1 or later. If not, send us an issue.

PR #251, opened 2018-12-18
Documentation & refactoring for custom kubectl from source

smukherj1:
Addresses #245

### DIFF
--- a/toolchains/kubectl/README.md
+++ b/toolchains/kubectl/README.md
@@ -24,9 +24,9 @@ go_rules_dependencies()
 go_register_toolchains()
 ```
 
-By default, the kubectl sources pulled in is defined [here](defaults.bzl).
-However, if you wanted to use release v1.13.1, call `kubectl_configure` as
-follows:
+By default, the kubectl sources and kubernetes repository infrastructure tools
+pulled in is defined [here](defaults.bzl). To use e.g., kubernetes release
+v1.13.1, call `kubectl_configure` as follows:
 
 ```python
 kubectl_configure(mame="k8s_config", build_srcs=True,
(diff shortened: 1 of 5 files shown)


toolchains/kubectl/kubectl_configure.bzl
```diff
@@ -74,21 +109,33 @@ def kubectl_configure(name, **kwargs):
...
-            sha256 = "21160531ea8a9a4001610223ad815622bf60671d308988c7057168a495a7e2e8",
-            strip_prefix = "repo-infra-b4bc4f1552c7fc1d4654753ca9b0e5e13883429f",
-            urls = ["https://github.com/kubernetes/repo-infra/archive/b4bc4f1552c7fc1d4654753ca9b0e5e13883429f.tar.gz"],
+            sha256 = _k8s_repo_tools_sha,
+            strip_prefix = "{}-{}".format(
+                _k8s_repo_tools_repo,
+                _k8s_repo_tools_commit
+            ),
+            urls = ["https://github.com/{}/{}/archive/{}.tar.gz".format(
+                _k8s_org,
```
nlopezgi: lets enable overriding these values as well

smukherj1:
@nlopezgi Unfortunately, with v1.10.0 I get 
```
E1218 16:38:19.395878       2 main.go:51] Error: Failed making a parser: unable to add directory "./vendor/k8s.io/apimachinery/pkg/util/sets/types": unable to import "./vendor/k8s.io/apimachinery/pkg/util/sets/types": cannot find package "./vendor/k8s.io/apimachinery/pkg/util/sets/types" in:
        /usr/local/google/home/suvanjan/.cache/bazel/_bazel_suvanjan/a1bcb3b63a242c66f0c6010821551247/sandbox/linux-sandbox/31/execroot/io_bazel_rules_k8s/vendor/k8s.io/apimachinery/pkg/util/sets/types
```

Seems to be because https://github.com/kubernetes/kubernetes/blob/fc32d2f3698e36b93322a3465f63a14e9f0eaead/vendor/k8s.io/apimachinery points to an invalid target?

Verified it works with v1.13.1 though. Addressed other comments. ptal

toolchains/kubectl/README.md
```diff
@@ -8,23 +8,74 @@ rule. At the moment, the tool's configuration does one of the following:
...
+    k8s_repo_tools_commit = "b4bc4f1552c7fc1d4654753ca9b0e5e13883429f",
+    k8s_repo_tools_sha256 = "21160531ea8a9a4001610223ad815622bf60671d308988c7057168a495a7e2e8",
+    k8s_repo_tools_prefix = "repo-infra-b4bc4f1552c7fc1d4654753ca9b0e5e13883429f"
+)
+```
+
+Note that by default the `k8s_repositories()` calls `kubectl_configure` if it
+hasn't already been called. This configures a `kubectl_toolchain` that looks for
+the `kubectl` tool in the sytem path. If no `kubectl` tool is found, trying to
+execute `bazel run` for any targets will not work.
```
nlopezgi: Also add note user must find out the right version of repository infrastructure tools that works with the version of kubectl sources they want to use by looking at the io_kubernetes_build target in https://github.com/kubernetes/kubernetes/blob/{k8s_commit}/build/root/WORKSPACE

smukherj1:
@nlopezgi Doc added. ptal

PR #257, opened 2019-01-14
Upgrade rules_proto

meteorcloudy:
To fix https://github.com/bazelbuild/bazel/issues/7113

PR #258, opened 2019-01-15
Update rules_proto (second try)

meteorcloudy:
https://github.com/bazelbuild/rules_k8s/pull/257 didn't fix the problem, this one does.


nlopezgi:
Note buildkite did pass but status was not reported: https://buildkite.com/bazel/rules-k8s-k8s/builds/464

PR #262, opened 2019-01-30
Update OWNERS

nlopezgi:


nlopezgi:
Matt passed ownership of the repo to us and Eric was helping while we ramped up. I'm happy to keep them in the OWNERS file if they still want to be there!

chrislovecnm:
@nlopezgi why don’t we give them a couple of days to respond, and if they do not, how about we remove them? How does that sound? Appreciate all the work both of them have done, and I know that both are not really active with the project.

nlopezgi:
sgtm

fejta:
SGTM

Issue #248, opened 2018-12-18
Fix Bazel incompatible changes


laurentlb:
I'd like to be able to build rules_closure with `--all_incompatible_changes`:
`bazel test --all_incompatible_changes //...`

It may not be possible to fix every issue now due to transitive dependencies, but we should try to fix the issues from this repository.

nlopezgi:
Hi @laurentlb ,
same as for rules_docker, happy to help you review any PRs you send to fix as many issues as possible in this repo.

terinjokes:
After fetching the fix from a0bbf543ac0eedec1ec699ca3d9b4e7bf1749aaf, I'm still running into an error when `--all_imcompatible_changes` is passed:

```
        File "/home/terin/.cache/bazel/_bazel_terin/2f6659f6d922032bbf4cf20ebfd5ecd5/external/io_bazel_rules_k8s/k8s/object.bzl", line 118, in ctx.runfiles
                list(ctx.attr.resolver.default_runfil...)
type 'depset' is not iterable. Use the `to_list()` method to get a list. Use --incompatible_depset_is_not_iterable=false to temporarily disable this check.
```

Looks like changing a handful of `list(.files)` to `.files.to_list()` resolved this error for me.

laurentlb:
There are still issues with these flags:
* `--incompatible_new_actions_api`
* `--incompatible_no_support_tools_in_action_inputs`
* `--incompatible_disable_deprecated_attr_params`
* `--incompatible_disallow_load_labels_to_cross_package_boundaries`
* `--incompatible_disallow_dict_plus`

laurentlb:
We plan to fix some of these flags soon in Bazel.

nlopezgi:
Thanks for the reminder, we'll schedule to make changes to fix next week.

laurentlb:
`--incompatible_disallow_dict_plus` is now enabled in Bazel at head.

smukherj1:
Going to work on `--incompatible_disallow_dict_plus`

smukherj1:
I tried to fix as many other issues I could find by trying to update external repositories for incompatible flags other than disallow_dict_plus. It seems some work is going to be needed in rules_grpc to fix some of the other errors. I wasn't able to update com_goole_protobuf because I get

```
[...]

PR #268, opened 2019-02-22
Fix --incompatible_disallow_dict_plus

smukherj1:
Also updated a few repositories to pull in fixes to other incompatible
changes that haven't been enabled by default in Bazel yet. More work is
needed in rules_grpc and com_google_protobuf. See #248

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -35,8 +35,8 @@ http_archive(
 
 http_archive(
     name = "io_bazel_rules_docker",
-    strip_prefix = "rules_docker-10ac85c3a91a38c6a0c178844c75b27092330ee9",
     sha256 = "ed9b4cd5d1c37e5b9243e4eaf31a8e155fd49f2a73955dc07512b5b476e63ea6",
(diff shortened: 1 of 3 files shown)


smukherj1:
I think the Bazel version being used by prow might be too old

WORKSPACE
```diff
@@ -146,9 +147,9 @@ go_register_toolchains()
 
 http_archive(
     name = "build_stack_rules_proto",
-    sha256 = "074dbfbeef97dfb2145cdda0bab6d1d51f459e90e0e835284a59b3880bbebd6e",
-    strip_prefix = "rules_proto-137014a36f389cfcb4987a567b7bd23a7a259cf9",
-    urls = ["https://github.com/stackb/rules_proto/archive/137014a36f389cfcb4987a567b7bd23a7a259cf9.tar.gz"],
+    strip_prefix = "rules_proto-f5d6eea6a4528bef3c1d3a44d486b51a214d61c2",
```
nlopezgi: buildifier will probably complain that sha needs to be before strip_prefix

smukherj1: Ran buildifier which fixed these

PR #273, opened 2019-03-06
Update rules_go to 0.17.0

### LINKED ISSUE #272: Get all tests passing with --incompatible_disable_legacy_cc_provider for Bazel 0.25.0
It seems that rules_k8s will start breaking with the next Bazel release 0.25 because of an incompatible change (https://github.com/bazelbuild/bazel/issues/7036)

`ERROR: /home/bazel/.cache/bazel/_bazel_bazel/16b7e4e760c79bffb5052db686d1c0f9/external/org_golang_x_sys/unix/BUILD.bazel:3:1: in libs attribute of _cgo_collect_info rule @org_golang_x_sys//unix:go_default_library%android_arm64%cgo_info: '@org_golang_x_sys//unix:go_default_library%android_arm64%cgo_c_lib' does not have mandatory providers: 'cc'. Since this rule was created by the macro 'go_library_macro', the error might have been caused by the macro implementation in /home/bazel/.cache/bazel/_bazel_bazel/16b7e4e760c79bffb5052db686d1c0f9/external/io_bazel_rules_go/go/private/rules/cgo.bzl:695:16`

I don't see any errors for rules_go, so it could be that rules_k8s is using an old version of rules_go?

smukherj1:
Fixes --incompatible_disable_legacy_cc_provider issues which fixes #272

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -130,12 +130,11 @@ py_library(
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "efa592f3f71a7e2f1cea98e431e2ccc8dba98375ae144f115fd4789f092869de",
-    strip_prefix = "rules_go-0.16.5",
-    urls = ["https://github.com/bazelbuild/rules_go/archive/0.16.5.tar.gz"],
+    sha256 = "6776d68ebb897625dead17ae510eac3d5f6342367327875210df44dbe2aeeb19",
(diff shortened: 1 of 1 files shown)


smukherj1:
Need to submit a separate buildifier fix like rules_docker

xingao267:
rules_go 0.18.0 is out

PR #274, opened 2019-03-06
Fix buildifier 0.22.0 lint warnings

smukherj1:
Disabled buildifier in buildkite and enabled it in e2e tests with
specific warnings disabled.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,5 +1,4 @@
 ---
-buildifier: true
 platforms:
   ubuntu1404:
     build_targets:
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,4 +1,6 @@
 ---
+# TODO (suvanjan): Re-enable once lint warnings have been fixed.
+# buildifier: true
 platforms:
   ubuntu1404:
     build_targets:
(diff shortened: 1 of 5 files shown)


.bazelci/presubmit.yml
```diff
@@ -1,5 +1,4 @@
 ---
-buildifier: true
```
nlopezgi: Leave a TODO to turn on again once we can disable specific warnings

test-e2e.sh
```diff
@@ -81,6 +81,18 @@ kubectl version
...
 
+# Install buildifier 0.22.0 and run lint checks
+wget -q https://github.com/bazelbuild/buildtools/releases/download/0.22.0/buildifier
+chmod +x ./buildifier
+
+# Check for issues with the format of our bazel config files.
+# All warnings from https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md
+# are enabled except:
+# rule-impl-return, uninitialized, return-value
+# TODO (suvanjan): reenable once issues are fixed
```
nlopezgi: reenable and move to buildkite...

.bazelci/presubmit.yml
smukherj1: Done

test-e2e.sh
smukherj1: Done

PR #287, opened 2019-03-11
rules_k8s: Support passing in a kubeconfig file

rohansingh:
Add support for passing in a generated kubeconfig file to `rules_k8s`
objects. This can be used to pass in a configuration that's generated by
Bazel.

### DIFF
--- a/k8s/apply.sh.tpl
+++ b/k8s/apply.sh.tpl
@@ -26,4 +26,5 @@ function exe() { echo "\$ ${@/eval/}" ; "$@" ; }
 RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
 
 PYTHON_RUNFILES=${RUNFILES} %{resolve_script} | \
-  exe  %{kubectl_tool} --cluster="%{cluster}" --context="%{context}" --user="%{user}" %{namespace_arg} apply $@ -f -
+  exe  %{kubectl_tool} --kubeconfig="%{kubeconfig}" --cluster="%{cluster}" \
+  --context="%{context}" --user="%{user}" %{namespace_arg} apply $@ -f -
--- a/k8s/create.sh.tpl
+++ b/k8s/create.sh.tpl
@@ -28,4 +28,5 @@ RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
 # TODO(mattmoor): Should we create namespaces that do not exist?
 
 PYTHON_RUNFILES=${RUNFILES} %{resolve_script} | \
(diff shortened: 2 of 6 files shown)


rohansingh:
This doesn't include any updates to relevant documentation. If this approach is acceptable, I'll make those updates as well.

k8s/apply.sh.tpl
```diff
@@ -26,4 +26,5 @@ function exe() { echo "\$ ${@/eval/}" ; "$@" ; }
 RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
 
 PYTHON_RUNFILES=${RUNFILES} %{resolve_script} | \
-  exe  %{kubectl_tool} --cluster="%{cluster}" --context="%{context}" --user="%{user}" %{namespace_arg} apply $@ -f -
+  exe  %{kubectl_tool} --kubeconfig="%{kubeconfig}" --cluster="%{cluster}" \
```
smukherj1: What does kubectl do when --kubeconfig is given a blank string? Would it be preferable to only specify --kubeconfig when there's something to specify? The if condition would move to the k8s_object rule.

k8s/create.sh.tpl
```diff
@@ -28,4 +28,5 @@ RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
 # TODO(mattmoor): Should we create namespaces that do not exist?
 
 PYTHON_RUNFILES=${RUNFILES} %{resolve_script} | \
-  exe %{kubectl_tool} --cluster="%{cluster}" --context="%{context}" --user="%{user}" %{namespace_arg} create $@ -f -
+  exe %{kubectl_tool} --kubeconfig="%{kubeconfig}" --cluster="%{cluster}" \
```
smukherj1: Same comment here as apply

k8s/apply.sh.tpl
rohansingh: When `--kubeconfig=''`, the behavior is the same as not specifying `--kubeconfig` at all. It's the same as `--cluster=''` and `--context=''` in that regard.

I agree that it would be cleaner and preferable to not output flags if the value is empty, since we're just getting lucky that the default value of all the current flags is an empty string. However, I think that should be a separate PR that would do this for all of the flags, rather than having different behavior for just this one flag.

rohansingh:
@smukherj1 Thanks for the prompt review! I'll add tests and docs (and fix the buildifier failure).

rohansingh:
Updated documentation and tests, and rebased. Note that I don't have an environment setup for running the e2e tests locally, so kind of shooting in the dark here.

PR #289, opened 2019-03-12
Update deps for --incompatible_remove_old_python_version_api

brandjon:
This moves the explicit rules_python dep up in the WORKSPACE file so it takes precedence. It also adds an explicit dep on subpar to override transitive deps on it that have yet to be updated.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -26,6 +26,15 @@ http_archive(
     url = "https://github.com/google/protobuf/archive/66dc42d891a4fc8e9190c524fd67961688a37bbe.tar.gz",
 )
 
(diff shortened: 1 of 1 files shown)


brandjon:
The error message indicates that your CI is using an older version of Bazel. This PR updates for an incompatible change flag added by Bazel 0.23, which will become flipped (i.e., actually breaking) in Bazel 0.25. The update requires features added by Bazel 0.23, hence why CI is failing.

smukherj1:
Alright I'll look into updating the Bazel version in our CI image. There's another renovate update PR that's failing because of this

PR #296, opened 2019-03-21
images is not required on k8s_object

jazzdan:
Unless I'm misunderstanding the meaning of required here? Let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,7 +540,7 @@ A rule for interacting with Kubernetes objects.
     <tr>
       <td><code>images</code></td>
       <td>
-        <p><code>string to label dictionary; required</code></p>
+        <p><code>string to label dictionary</code></p>
         <p>When this target is <code>bazel run</code> the images
           referenced by label will be published to the tag key.</p>
        <p>The published digests of these images will be substituted


erain:
/ok-to-test

You are right images is only required for *some* k8s resources, e.g. deployment or pod. But it is quite common though to have rules_docker to build an image and rules_k8s to deploy an image. 

Could you help also double check the rule implementation if the attribute is set as required in the code?

jazzdan:
Hi @erain, that's great context I wasn't aware of. I'm using this on a [deployment with no images](https://github.com/windmilleng/servantes/blob/759e02bdfba2a5591dfbca1463f5e102e83f9ca0/BUILD) so that I can generate YAML _without_ building an image and it seems to work just fine.

I'm not really bazel fluent (just started learning it last week) but from my reading of [the code](https://github.com/bazelbuild/rules_k8s/blob/master/k8s/object.bzl#L53) it looks to me like `images` is always optional?

smukherj1:
The images attribute is not mandatory as per the declaration [here](https://github.com/bazelbuild/rules_k8s/blob/8bcee3c0ab3a860010fb1f817da8785ace8f5d25/k8s/object.bzl#L316).

That being said, this is blocked because our CI seems to have an issue as explained [here](https://github.com/bazelbuild/rules_k8s/pull/294#issuecomment-476194139). I'm going to try and workaround this flake to unblock this PR

PR #300, opened 2019-03-29
Return providers instead of struct from rule impl functions

alex1545:
This addresses the incompatible change via the `--incompatible_disallow_struct_provider_syntax` flag described [here](https://github.com/bazelbuild/bazel/issues/7347).

### DIFF
--- a/k8s/object.bzl
+++ b/k8s/object.bzl
@@ -122,10 +122,13 @@ def _impl(ctx):
 
     return [
         DefaultInfo(
-            runfiles = ctx.runfiles(files = [
-                ctx.executable.resolver,
-                ctx.outputs.substituted,
-            ] + list(ctx.attr.resolver[DefaultInfo].default_runfiles.files) + all_inputs),
+            runfiles = ctx.runfiles(
+                files = [
+                    ctx.executable.resolver,
+                    ctx.outputs.substituted,
+                ],
+                transitive_files = list(ctx.attr.resolver[DefaultInfo].default_runfiles.files) + all_inputs,
+            ),
         ),
     ]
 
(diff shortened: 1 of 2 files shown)


k8s/object.bzl
```diff
@@ -120,10 +120,14 @@ def _impl(ctx):
...
-    return struct(runfiles = ctx.runfiles(files = [
-        ctx.executable.resolver,
-        ctx.outputs.substituted,
-    ] + list(ctx.attr.resolver.default_runfiles.files) + all_inputs))
+    return [
+        DefaultInfo(
+            runfiles = ctx.runfiles(files = [
+                ctx.executable.resolver,
+                ctx.outputs.substituted,
+            ] + list(ctx.attr.resolver.default_runfiles.files) + all_inputs),
```
xingao267: ctx.attr.resolver.default_runfiles.files should be accessed in the new provider way as well. i.e. ctx.attr.resolver[DefaultInfo].default_runfiles.files

xingao267: Also use transitive_files attribute like this https://github.com/bazelbuild/rules_docker/pull/724/files#diff-0ebb096ed27e3e315a65e0f87e1a5bf9R53

k8s/object.bzl
```diff
@@ -274,10 +282,14 @@ def _reverse(ctx):
...
-    return struct(runfiles = ctx.runfiles(files = [
-        ctx.executable.reverser,
-        ctx.file.template,
-    ] + list(ctx.attr.reverser.default_runfiles.files)))
+    return [
+        DefaultInfo(
+            runfiles = ctx.runfiles(files = [
+                ctx.executable.reverser,
+                ctx.file.template,
+            ] + list(ctx.attr.reverser.default_runfiles.files)),
```
xingao267: Similar as above

k8s/object.bzl
alex1545: Done

k8s/object.bzl
alex1545: Done

PR #327, opened 2019-05-02
Allow kubeconfig in k8s_defaults and .delete rules

steeve:
It is now possible to use `kubeconfig` in `k8s_defaults`, which we use with k3s internally.
In `.delete` rules, `kubeconfig` was ignored, so fix that.

### DIFF
--- a/examples/hellogrpc/e2e-test.sh
+++ b/examples/hellogrpc/e2e-test.sh
@@ -114,9 +114,11 @@ delete() {
 }
 
 check_kubeconfig_args() {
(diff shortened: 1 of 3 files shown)


smukherj1:
Hi @steeve, Thanks for you PR and sorry for the delayed response. Could you please add a test for this [here](https://github.com/bazelbuild/rules_k8s/blob/8bcee3c0ab3a860010fb1f817da8785ace8f5d25/examples/hellogrpc/e2e-test.sh#L116) ensuring the kubeconfig argument is validated for ".delete" in addition to ".apply". Thanks!

steeve:
Sure.
Amended and rebased.

PR #337, opened 2019-05-17
Fixing kubectl toolchain docs

chrislovecnm:
The docs do not give the correct path.  They give the path inside our
repo, not the path you need to use in your workspace file.


chrislovecnm:
@nlopezgi do we have flakes still?  This is a doc update that is failing

chrislovecnm:
Filed https://github.com/bazelbuild/rules_k8s/issues/341

PR #342, opened 2019-05-21
Add support for select statement for k8s_objects.

Globegitter:
We are making use of `--define` flags to deploy targets into different environments. Right now we are making use of an `alias` to achieve this but it has limitations and adds code complexity that we would like to remove. Unfortunately macros can not manipulate `select` statements directly so I have to treat a `dict` as a special case. See https://github.com/bazelbuild/bazel/issues/8419

### DIFF
--- a/k8s/objects.bzl
+++ b/k8s/objects.bzl
@@ -58,13 +58,13 @@ _run_all = rule(
     implementation = _run_all_impl,
 )
 
-def _reverse(lis, reverse=False):
+def _reverse(lis, reverse = False):
     if reverse:
         return reversed(lis)
     else:
         return lis
 
-def _cmd_objects(cmd, objects, reverse=False):
+def _cmd_objects(cmd, objects, reverse = False):
     if type(objects) == "dict":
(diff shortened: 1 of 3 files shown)


chrislovecnm:
/ok-to-test

Looks good, a rebase would be awesome :)

Globegitter:
@chrislovecnm rebased :)

k8s/objects.bzl
```diff
@@ -58,20 +58,31 @@ _run_all = rule(
     implementation = _run_all_impl,
 )
 
+def _reverse(lis, reverse = False):
+    if reverse:
+        return reversed(lis)
+    else:
+        return lis
+
+def _cmd_objects(cmd, objects, reverse = False):
```
nlopezgi: add short doc to this and def above

Globegitter: Done.

Globegitter:
Anything still blocking this to get this merged?

nlopezgi:
apilogies for the delay, thanks for the poke

PR #344, opened 2019-05-22
Add a namespace garbage collector.

smukherj1:
Best effort deletion of kubernetes namespaces created by e2e test runs
more than an hour old.


examples/e2e_gc.py
```diff
@@ -0,0 +1,79 @@
...
+import logging
+import json
+from datetime import datetime
+
+# Logger instance
+_log = None
+
+def list_namespaces():
+    """
+    Returns a list of strings of namespace names that were created the e2e tests
```
nlopezgi: by the

smukherj1: Done

PR #353, opened 2019-05-30
Update dependency on grpc

laurentlb:
With this change, the repository can build with Bazel flag
--incompatible_depset_is_not_iterable, which will be enabled by default
soon.


laurentlb:
(cc @vladmos)

PR #357, opened 2019-06-04
Update protobuf pin

alex1545:
Renovate bot fails to update the protobuf dependency (#356) as it's unable to guess the new `strip_prefix` value.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -18,9 +18,9 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file"
 
 http_archive(
     name = "com_google_protobuf",
-    sha256 = "f1748989842b46fa208b2a6e4e2785133cfcc3e4d43c17fecb023733f0f5443f",
-    strip_prefix = "protobuf-3.7.1",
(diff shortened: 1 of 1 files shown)


achew22:
FYI, the RenovateBot team has been very receptive to bug reports in the past over this kind of stuff. I filed https://github.com/renovatebot/renovate/issues/3888 for you to track Renovate being able to handle this correctly.

alex1545:
Thank you very much @achew22 for your immediate interaction regarding this issue.
But I intentionally haven't filed an issue as I'm not sure how this type of scenario can be handled properly. The `strip_prefix` can actually be anything you want and doesn't need to be 'similar' in any way to the commit reference (commit sha or tag). So I thought that the current approach of guessing that the `strip_prefix` ends with the commit reference is probably good enough.

achew22:
> The strip_prefix can actually be anything you want and doesn't need to be 'similar' in any way to the commit reference (commit sha or tag). So I thought that the current approach of guessing that the strip_prefix ends with the commit reference is probably good enough.

That's a good bit of clarity. Asking for the `strip_prefix` to be updated is a github specific thing. In the case of a commit it would be the commit hash and in the case of a tag it would be the tag name.

PR #427, opened 2019-10-03
Add in a no-op Go resolver

smukherj1:
Had to refactor the dependencies a lot. Will need to cut a release and
then update the README with the new instructions.


k8s/go/cmd/resolver/BUILD
```diff
@@ -0,0 +1,18 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
```
nlopezgi: add licence header + visiblity

k8s/go/cmd/resolver/resolver.go
```diff
@@ -0,0 +1,39 @@
+package main
```
nlopezgi: license header

k8s/go/cmd/resolver/BUILD
smukherj1: Done

k8s/go/cmd/resolver/resolver.go
smukherj1: Done

Issue #433, opened 2019-10-09
rules_k8s should use go-containerregistry


smukherj1:
[//k8s:resolver.py](https://github.com/bazelbuild/rules_k8s/blob/ddf25337bfff860a01040a32d85ccbf5195a1f52/k8s/resolver.py) uses the python containerregistry library which is no longer maintained. We should migrate to go-containerregistry.

nlopezgi:
related to https://github.com/bazelbuild/rules_docker/issues/580

fejta:
/sub
https://github.com/kubernetes/test-infra/issues/14678

mariusgrigoriu:
Also related to #378

smukherj1:
#435 finished implementing the resolver in Go and switched rules_k8s to use the go resolver by default. We'll need people to try out the latest ruleS_k8s commit and let us know if there are any issues with the Go resolver. We added a new attribute `use_legacy_resolver` to the `k8s_object` which defaults to False. When set to true, `k8s_object` will use the new Go resolver.

fejta:
See https://github.com/kubernetes/test-infra/issues/14763 we're needing to use the use_legacy_resolver in at least one place due to a quay.io image that fails to resolve.

fejta:
Just tried at 1b5dcb210548b68ce1d215f156436e102df3ba73 and this doesn't yet appear to be working:


https://github.com/kubernetes/test-infra/blob/015d9914337ac54f0b1803364d85b141f55206bb/prow/cluster/monitoring/BUILD.bazel#L13

```console
# with use_legacy_resolver=True
fejta@fejta3:~/src/gh/test-infra$ bazel run //prow/cluster/monitoring:prometheus_operator
INFO: Found 1 target...
Target //prow/cluster/monitoring:prometheus_operator up-to-date:
  bazel-bin/prow/cluster/monitoring/prometheus_operator
INFO: Elapsed time: 2.069s, Critical Path: 0.19s
INFO: 0 processes.
INFO: Build completed successfully, 4 total actions
INFO: Build completed successfully, 4 total actions
apiVersion: apps/v1
kind: Deployment
metadata:
  labels: {app: prometheus-operator}
  name: prometheus-operator
  namespace: prow-monitoring
spec:
  replicas: 1
  selector:
    matchLabels: {app: prometheus-operator}
  template:
    metadata:
      labels: {app: prometheus-operator}
    spec:
      containers:
[...]

PR #444, opened 2019-10-17
Support resolving YAML streams in Go resolver

### LINKED ISSUE #443: Regression: go resolver does not retain multiple documents found in an yaml file
The new go resolver does not return the same output as the python resolver when the template has multiple yaml files. Eg A .yaml file with --- separating different yaml objects. 

This is because https://godoc.org/gopkg.in/yaml.v2#Unmarshal only parses the first document in the input and the remaining objects are skipped by [resolveYAML](https://github.com/bazelbuild/rules_k8s/blob/master/k8s/go/cmd/resolver/resolver.go#L291).

I'm attaching the file I happen to be testing with, but I don't think it's content specific.
[crds.yaml.txt](https://github.com/bazelbuild/rules_k8s/files/3737202/crds.yaml.txt)

smukherj1:
This fixes #443

### DIFF
--- a/k8s/go/cmd/resolver/resolver_test.go
+++ b/k8s/go/cmd/resolver/resolver_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"testing"
 
@@ -36,7 +37,7 @@ val8:
 	r := &resolver{
 		strResolver: sr,
 	}
-	if _, err := r.resolveYAML(y); err != nil {
+	if _, err := r.resolveYAML(bytes.NewBuffer(y)); err != nil {
 		t.Fatalf("Failed to resolve YAML: %v", err)
 	}
 	want := make(map[string]bool)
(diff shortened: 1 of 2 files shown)


smukherj1:
Relevant for #433

k8s/go/cmd/resolver/resolver_test.go
```diff
@@ -48,3 +49,105 @@ val8:
...
+func TestYAMLStream(t *testing.T) {
+	testCases := []struct {
+		name string
+		// yamlDocs is a string with one or more YAML documents.
+		yamlDocs string
+		// want is the list of strings the YAML walker should visit.
+		want []string
+		// wantDocs is the number of docs the resolver should have operated on.
+		wantDocs int
+	}{
```
alex1545: Not sure if it's a valid scenario, but can we have an empty yaml file? If so, can we have a test for that too?
```
- val1
- val2
---
---
- val3
- val4
```

PR #446, opened 2019-10-18
Go resolver should return strings as is on any kind of error

smukherj1:
This is how the python resolver behaves. Tested with
bazel run //prow/cluster/monitoring:prometheus_operator on
kubernetes/test-infra

### DIFF
--- a/k8s/go/cmd/resolver/resolver.go
+++ b/k8s/go/cmd/resolver/resolver.go
@@ -233,11 +233,11 @@ func resolveString(r *resolver, s string) (string, error) {
 	}
 	auth, err := authn.DefaultKeychain.Resolve(t.Context())
 	if err != nil {
-		return "", fmt.Errorf("unable to get the authenticator using the default keychain for image %v: %v", t, auth)
+		return s, nil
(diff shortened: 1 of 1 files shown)


k8s/go/cmd/resolver/resolver.go
```diff
@@ -233,11 +233,11 @@ func resolveString(r *resolver, s string) (string, error) {
 	}
 	auth, err := authn.DefaultKeychain.Resolve(t.Context())
 	if err != nil {
-		return "", fmt.Errorf("unable to get the authenticator using the default keychain for image %v: %v", t, auth)
+		return s, nil
```
alex1545: Can you also include a short explanation in the function description comment why the returned error needs to be always nil.

smukherj1:
Another fix for #433

k8s/go/cmd/resolver/resolver.go
smukherj1: Ack. Will address in follow up PR as prow CI has been running for 10mins already it takes ~30mins

PR #462, opened 2019-10-31
Remove PY2 notice from the README

zegl:
This is no longer necessary since the resolver was migrated (released in 0.3.0), and #461 (not yet released).

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,17 +41,6 @@ load("@io_bazel_rules_k8s//k8s:k8s_go_deps.bzl", k8s_go_deps = "deps")
 k8s_go_deps()
 ```
 
-NEW: Starting with Bazel 0.27.0, you also need to add to your .bazelrc
-file the following:
-
-```
-build --host_force_python=PY2
-test --host_force_python=PY2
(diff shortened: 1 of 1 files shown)


zegl:
You might want to wait with merging this PR until the next release is made, to not confuse anyone reading the readme on master.

terinjokes:
By removing this section, it implies that stamper should work under Python 2 and Python 3 without additional configuration. However, without this it seems Bazel complains about Python versions and ends the build:

```
Error occurred while attempting to use the default Python toolchain (@rules_python//python:autodetecting_toolchain).
According to '/usr/bin/python -V', version is 'Python 2.7.13', but we need version 3. PATH is:

/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

Please ensure an interpreter with version 3 is available on this platform as 'python3' or 'python', or else register an appropriate Python toolchain as per the documentation for py_runtime_pair (https://github.com/bazelbuild/rules_python/blob/master/docs/python.md#py_runtime_pair).
```

Is there something else I'm supposed to be doing?

Issue #351, opened 2019-05-28
Support stamping substitutions


christian-roggia:
This issue is somehow related to https://github.com/bazelbuild/rules_k8s/issues/266.

It would be nice if the following example was possible:
```
k8s_object(
    ...
    substitutions = {
        "%{expand_template_variable}": "{stamp_variable}",
    },
    ...
)
```

This is especially useful if you want to embed information inside the YAML that is coming from the workspace.

nlopezgi:
Hi, I'd be happy to review any PRs you want to send to support this.

PR #477, opened 2019-11-29
Resolve stamped variables in Kubernetes objects

zoidbergwill:
I was looking at this as work towards fixing https://github.com/bazelbuild/rules_k8s/issues/351

### DIFF
--- a/k8s/go/cmd/resolver/resolver.go
+++ b/k8s/go/cmd/resolver/resolver.go
@@ -206,6 +206,8 @@ type resolver struct {
 	// numDocs stores the number of documents the resolver worked on when
 	// resolveYAML was called. This is used for testing only.
 	numDocs int
+	// stamper resolves stuff
+	stamper *compat.Stamper
 }
 
 // resolveString resolves a string found in the k8s YAML template by replacing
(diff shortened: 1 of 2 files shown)


zoidbergwill:
I got a bit stuck, thinking about how to send the dict of substitutions to the resolver.go to use the stamp-info-file when resolving the substitutions.

I could figure out how to do it in Python, but not Starlark in a rule.

rohansingh:
FYI, I took a shot at this in #505. In terms of this:

> I got a bit stuck, thinking about how to send the dict of substitutions to the resolver.go to use the stamp-info-file when resolving the substitutions.
>
> I could figure out how to do it in Python, but not Starlark in a rule.

I think this is a pretty good idiom for when you want to pass a somewhat complicated structure from Starlark to a binary:

```python
ctx.actions.write(
        output = some_file,
        content = struct(
            key = some_dict,
        ).to_json(),
    )
```

The binary can then read `some_file` and parse it as JSON.

zoidbergwill:
@rohansingh That's a great solution. Will take a look at revisiting this.

fejta:
Please holler when this is ready for review!

fejta:
Please reopen if you get the chance to update it. Closing until then.

PR #517, opened 2020-02-03
Update rules_go

meteorcloudy:
Fixes bazelbuild/bazel#10644

### DIFF
--- a/k8s/k8s.bzl
+++ b/k8s/k8s.bzl
@@ -62,11 +62,12 @@ py_library(
     if "io_bazel_rules_go" not in excludes:
         http_archive(
             name = "io_bazel_rules_go",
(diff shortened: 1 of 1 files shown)


nlopezgi:
Hi @meteorcloudy , we have https://github.com/bazelbuild/rules_k8s/pull/516 also pending which is updating the pin to v0.21.1. If we pin to a non release commit, I'm unsure if renovate bot will continue to do its job properly. I see that the [commit](https://github.com/bazelbuild/rules_go/commit/5474789768ede024a294137d8f249d5cb5b60627) you pin to  is not yet in a release. Would it be possible to do a release with this commit if its necessary to fix Bazel CI?

nlopezgi:
replaced by https://github.com/bazelbuild/rules_k8s/pull/519

Issue #536, opened 2020-03-06
Dependency on volatile-status.txt breaks remote cache


mariusgrigoriu:
The [Bazel docs](https://docs.bazel.build/versions/master/user-manual.html#workspace_status) state:
> Bazel pretends that the volatile file never changes

However, that is [untrue when it comes to the remote cache](https://github.com/bazelbuild/bazel/issues/10075).

Our CI environment runs e2e tests that depend on a `k8s_object` on every run even when there are no changes due to the interaction between rules_k8s, stamping, and the remote_cache.

I'd like to propose dropping the dependency on the volatile file, while retaining the stable file, or at least making it optional as a workaround since the Bazel maintainers are pushing back on fixing the underlying issue.

mariusgrigoriu:
I found that depending on the `.substituted.yaml` output file is a workaround to this remote caching issue. The downside is that there may be cases when the test doesn't run due to some changes in the rule related to stamping.

fejta:
General best practice here is to avoid stamping -- stamping is by definition volatile.

Another practice we use over in kubernetes/test-infra is to separate out the stamp layer into its own final layer.

AKA instead of making the final stamped layer include a large go binary in it, we put this into an intermediary layer.

https://github.com/kubernetes/test-infra/blob/35cfa7b7d66098fabedb8fb1c0e15be29088971e/prow/def.bzl#L26-L49

In other words instead of using the rules_docker container_image directly we wrap this into a prow_image macro which separates out the stamped layer from the large binary layer.

This way 99% of the image layers remain reproducible (and therefore cached) and just the tiny final layer with stamp info changes between builds (assuming the image hasn't changed)

mmikitka:
@mariusgrigoriu and @fejta I created https://github.com/bazelbuild/rules_k8s/pull/592 as another attempt at addressing this issue. It exposes the `stamp_srcs` attribute as a means to override `volatile-status.txt` (and `stable-status.txt` if desired).

mariusgrigoriu:
That looks like a creative solution! 

We've settled into adding the `.substituted.yaml` output and a custom resolver as data dependencies on e2e tests for our project. The test uses those dependencies only to trigger e2e execution. The custom resolver substitutes values like cluster and namespace, which are different per pipeline run so we don't run e2e with every commit.

[...]

PR #537, opened 2020-03-06
drop volatile status dependency

mariusgrigoriu:
PR that might address #536 

This PR drops the dependency on the volatile status, which causes k8s_object and downstream dependencies to rebuild unnecessarily when using remote cache.

### DIFF
--- a/k8s/object.bzl
+++ b/k8s/object.bzl
@@ -70,7 +70,7 @@ def _impl(ctx):
             image_spec = {"name": resolved_tag}
             if image.get("legacy"):
                 image_spec["tarball"] = _runfiles(ctx, image["legacy"])
-                all_inputs += [image["legacy"]]
+                all_inputs.append(image["legacy"])
 
             blobsums = image.get("blobsum", [])
             image_spec["digest"] = ",".join([_runfiles(ctx, f) for f in blobsums])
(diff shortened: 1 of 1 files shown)


mariusgrigoriu:
Apologies about the `+=` change. Seems that buildifier auto-formatter likes `.append` better for some reason. 

I mostly wanted to have a discussion about disabling volatile status stamping to work around https://github.com/bazelbuild/bazel/issues/10075 and whether this is an appropriate workaround or whether there is a better way.

fejta:
Thanks for this, can you rebase?

/assign
/ok-to-test

> why change += to .append ?

Bazel will eventually require this

k8s/object.bzl
```diff
@@ -89,17 +89,17 @@ def _impl(ctx):
...
-            image_specs += ["';'".join([
+            image_specs.append("';'".join([
                 "%s=%s" % (k, v)
                 for (k, v) in image_spec.items()
-            ])]
+            ]))
 
     # Add workspace_status_command files to the args that are pushed to the resolver and adds the
     # files to the runfiles so they are available to the resolver executable.
-    stamp_inputs = [ctx.info_file, ctx.version_file]
```
fejta, line 102: Strange, I cannot see how this is even used?

mariusgrigoriu:
Looks like we would need to change https://github.com/bazelbuild/rules_k8s/blob/master/print-workspace-status.sh#L22 to `STABLE_E2E_NAMESPACE` and any other references to that stamp. Do you still need this rebased?

fejta:
`E2E_NAMESPACE` is volatile -- it changes every build/PR -- why should this be changed to `STABLE_E2E_NAMESPACE`?

mariusgrigoriu:
This PR attempts to work around the issue that volatile stamps trigger rebuilds when a remote cache is used by dropping the dependency on the volatile file. The CI job fails since volatile stamps are no longer available to the rule.

So is this really acceptable after all?

fejta:
This repo isn't using a remote cache right now.

PR #541, opened 2020-03-30
fix encoding error

asiekkowa:
Fix error "UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 12345: ordinal not in range(128)"


asiekkowa:
@smukherj1 could you please merge it?

PR #546, opened 2020-04-09
Add descriptions on `resolver_args`

ymotongpoo:
In #161 and #162, the attribution `resolver_args` was added to `k8s_object`, but the attribute was not documented. As described in the doc, some options such as `--allow_unused_images` should be passed via this option.


alex1545:
Thank you @ymotongpoo for improving our docs with your recent contributions.

Issue #562, opened 2020-05-26
Substitutions in k8s_object does not work. :(


asiekkowa:
Substitutions in k8s_object does not work. :(

Tested yaml file:
```yaml
kind: %{RESOURCE_TYPE}                                                                                                                                                   
apiVersion: argoproj.io/v1alpha1
metadata:
  name: test
  generateName: test-e2e
spec:
  arguments:
    parameters:
...
```

BUILD.bazel file:
```
load("@io_bazel_rules_k8s//k8s:object.bzl", "k8s_object")                                                                                                                
  
images = {
...
}

k8s_object(
    name = "workflow.template",
    kind = "workflowtemplate",
    template = "workflow.template.yaml",
    images = images,
    substitutions = {
        "%{RESOURCE_TYPE}": "WorkflowTemplate",
    },
    cluster = "XYZ",
)
k8s_object(
    name = "workflow",
    kind = "workflow",
    template = "workflow.template.yaml",
    images = images,
    substitutions = {
        "%{RESOURCE_TYPE}": "Workflow",
    },
    cluster = "XYZ",
)
```

Output of command
```
bazel run :workflow
``` 
is correct, variable RESOURCE_TYPE is being replaced.

Unfortunately when I try to run:
```
bazel run :workflow.delete #(or .create)
``` 
I got:
```
  File "/home/abc/.cache/bazel/_bazel_abc/c27b7a4375a4d9fe604f9c7388818c81/execroot/__main__/bazel-out/k8-fastbuild/bin/workflow.delete.runfiles/com_github_yaml_pyyaml/lib3/yaml/scanner.py", line 260, in fetch_more_tokens
    self.get_mark())
yaml.scanner.ScannerError: while scanning for the next token
found character '%' that cannot start any token
  in "<unicode string>", line 1, column 7:
    kind: %{RESOURCE_TYPE}
```

tested bazel versions:
- 1.2.1
- 3.1.0

diegohavenstein:
This is exactly the same issue I described in #556 ...

20 days ago, haven't heard from any maintainer of the project yet. We may have to fix this ourselves it seems

r2r-dev:
Hi,

I've managed to fix it by passing substituted file to reverser file instead of plain template as it is now. I'll fill in a PR today.

diegohavenstein:
Thanks @r2r-dev !

If you do, could you please ref #556 as well? Thanks!

asiekkowa:
[...]

Issue #556, opened 2020-05-12
Unable to use delete script when substitution done in the name of K8s object


diegohavenstein:
I am having issues with the `.delete` script on a K8s job using a substitution as part of its name.

The issue can be reproduced with other K8s abstractions (e.g. deployment). Here is a minimum example to reproduce the issue:

1. Put the files below in a dir and run `bazel run jobfile_k8s_objects.create`.
2. You'll see the `pi-whatever` job running to completion, as we are using `name: pi-{job-name}` and `substitutions = {"{job-name}" : "whatever"}`.
3. Try to remove the job again by `bazel run jobfile_k8s_objects.delete`. This will not remove the job. It will also not fail (due to `--ignore-not-found=true`?)

This problem does NOT arise if no substitutions are part of the object name, e.g. if we used instead `name: pi` and `substitutions = {}`

Please let me know if I am doing something wrong here, or whether this is a known issue.

BUILD:
```
load(":job.bzl", "k8s_config")
config = k8s_config(name = "jobfile")
```

job.bzl:
```
load("@io_bazel_rules_k8s//k8s:object.bzl", "k8s_object")
load("@io_bazel_rules_k8s//k8s:objects.bzl", "k8s_objects")

def k8s_config(name):
    k8s_object(name=name + "_k8s_object",
               kind = "job",
               template = "job.yaml",
               substitutions = {"{job-name}" : "whatever"},
               cluster = "minikube",
               context = "minikube",
               namespace = "default",
               image_chroot = "",
               images = {})
    k8s_objects_name = name + "_k8s_objects"
    k8s_objects(name=k8s_objects_name,
                objects = [name + "_k8s_object"])
```

job.yaml:
```
apiVersion: batch/v1
kind: Job
metadata:
  name: pi-{job-name}
spec:
  template:
    spec:
      containers:
      - name: pi
        image: perl
        command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
      restartPolicy: Never
  backoffLimit: 4
```

WORKSPACE:
```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "io_bazel_rules_k8s",
    sha256 = "cc75cf0d86312e1327d226e980efd3599704e01099b58b3c2fc4efe5e321fcd9",
    strip_prefix = "rules_k8s-0.3.1",
[...]

PR #565, opened 2020-06-03
Use substituted template for delete action

r2r-dev:
Use substituted file instead of plain template for "reversed" k8s object, thus enabling its correct handling for ".delete" targets.

Fixes: #562 #556


r2r-dev:
@googlebot I signed it!

r2r-dev:
@googlebot I signed it!

AleksanderGondek:
Nice.

PR #570, opened 2020-06-19
Refactor e2e testing

fejta:
Primarily to make the logs easier to read and understand what failed.

Also:
* Remove some unnecessary .bazelrc customization
* Switch to py3
* Make cc hellogrpc client fail when the rpc fails
* Mac support for edit.sh

Finally:
* Turn off py, cc hellogrpc coverage
  - Should probably just delete this the internals a container_image are irrelevant to rules_k8s
  - Issues with glibc .so, so just disable them for now (added benefit: tests are faster)


examples/hellogrpc/e2e-test.sh
```diff
@@ -18,156 +16,189 @@ set -e
...
+            echo "$ip"
+            break
+        fi
+        echo -n .
+        sleep 5
+    done
+
+    local output
+    local tmp_exec=hellogrpc_$1_client
+    logfail bazel run "//examples/hellogrpc/$1/client" "--script_path=$tmp_exec"
```
spiffxp, line 88:
```suggestion
    # Make Bazel generate a temporary script that runs the client executable
    # This will only generate the temp executable. It won't actually run it.
    local output
    local tmp_exec=hellogrpc_$1_client
    logfail bazel run "//examples/hellogrpc/$1/client" "--script_path=$tmp_exec"
```

fejta:
This needs https://github.com/kubernetes/test-infra/pull/18023

examples/hellogrpc/e2e-test.sh
```diff
@@ -18,156 +16,189 @@ set -e
...
+
+    echo -n "$tmp_exec: "
+    for i in {1..5}; do
```
spiffxp, line 91:
```suggestion
    # the service may not be up immediately after ip allocation so retry a few times
    for i in {1..5}; do
```

examples/hellogrpc/e2e-test.sh
```diff
@@ -18,156 +16,189 @@ set -e
...
+        sleep 10
+    done
+
+    if [[ -z "$output" ]]; then
+        output=$(./${tmp_exec} "$ip")
+    fi
+
+
+    # Make Bazel generate a temporary script that runs the client executable.
+    # This will only generate the temp executable. It won't actually run it.
```
spiffxp, line 106: suggested moving this to the `bazel run` call above

PR #577, opened 2020-06-24
Drop cc examples.

fejta:
The rules_k8s repo will deploy resources to kubernetes via the apiserver,
optionally calling out to rules_docker to replace the declared image
with a locally built one.

All the examples in various languages are totally irrelevant to this repo
and add a lot of development drag and churn keeping the dependencies up to date.

AKA so long the value of an image is a rules_docker target it will work.
The rules_k8s stuff neither knows nor cares how this image was created and
what languages it uses internally.

/assign @michelle192837 @chases2


fejta:
Woops deleted a VERY necessary `shift`, stamp again?

PR #582, opened 2020-07-08
Add a ".resolve" alias for the resolve task

dfinninger:
When automating interactions with the outputs of a k8s_object, it's easy to find and operate on tasks like ".apply" because there is a sigil (literally, ".apply") that you can search for in the target output. Because there is no corresponding sigil for the "resolve" target it's difficult to filter for this task.

Looking for the absence of "apply", "delete", etc is not very future-proof, since other target types could be introduced. Filtering out targets that contain a period between the colon and the newline is restrictive on target names.

I believe this is would achieve the desired effect.

### DIFF
--- a/k8s/objects.bzl
+++ b/k8s/objects.bzl
@@ -95,6 +95,7 @@ def k8s_objects(name, objects, **kwargs):
     # TODO(mattmoor): We may have to normalize the labels that come
     # in through objects.
     _run_all(name = name, objects = _cmd_objects("", objects), delimiter = "echo ---\n", **kwargs)
+    _run_all(name = name + ".resolve", objects = _cmd_objects("", objects), delimiter = "echo ---\n", **kwargs)
     _run_all(name = name + ".create", objects = _cmd_objects(".create", objects), **kwargs)
     _run_all(name = name + ".delete", objects = _cmd_objects(".delete", objects, True), **kwargs)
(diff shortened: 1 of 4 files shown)


k8s/objects.bzl
```diff
@@ -95,6 +95,7 @@ def k8s_objects(name, objects, **kwargs):
     # TODO(mattmoor): We may have to normalize the labels that come
     # in through objects.
     _run_all(name = name, objects = _cmd_objects("", objects), delimiter = "echo ---\n", **kwargs)
+    _run_all(name = name + ".resolve", objects = _cmd_objects("", objects), delimiter = "echo ---\n", **kwargs)
```
fejta, line 98: Please add test coverage for this target

fejta:
Please do add the test coverage!

dfinninger:
Oh, I didn't see the email until just now! Thanks for looking at this.

I can't see where in this repo to add the test for the `.resolve` target, since it's piggy-backing on the definition just above it. If you could point me to the location of the tests I'd be happy to add them!

fejta:
via bazel-bin https://github.com/bazelbuild/rules_k8s/blob/master/examples/hellogrpc/e2e-test.sh
via bazel run https://github.com/bazelbuild/rules_k8s/blob/master/examples/hellohttp/e2e-test.sh

PR #410, opened 2019-08-09
Add .diff command

samschlegel:
Now that `diff` is released, this adds a `.diff` command. In order to avoid pushing when resolving for the purposes of the diff, this also adds a `no_push` flag to `resolver.py`

Fixes #242

PR #584, opened 2020-07-10
Add .diff target

### LINKED ISSUE #242: bazel run //path/to:target.diff command
Maybe it is possible and I missed it, but is it possible to do a `.diff` like you would a `.apply`? I'm trying to onboard my free standing YAML configuration into rules_k8s and I believe it will be the same, but it would be nice to know what happens if I were to apply the change in production. Is there anything like that available?

I looked in the docs on this project and also in the docs for `kubectl` to see if I could make the apply command do a diff. It looks like you can get it to do a `--dry-run` but that is entirely local.

samschlegel:
Previously #410

This adds a `.diff` target to `k8s_object`. In order to avoid pushing when resolving for the purposes of the diff, this also adds a no_push flag to the resolver.

Fixes #242

### DIFF
--- a/examples/hellogrpc/e2e-test.sh
+++ b/examples/hellogrpc/e2e-test.sh
@@ -121,6 +121,11 @@ edit() {
     logfail "./examples/hellogrpc/$1/server/edit.sh" "$2"
 }
 
+diff() {
+    logfail bazel build "examples/hellogrpc/$1/server:staging.diff"
+    logfail "bazel-bin/examples/hellogrpc/$1/server/staging.diff"
+}
+
 update() {
     logfail bazel build "examples/hellogrpc/$1/server:staging.replace"
     logfail "bazel-bin/examples/hellogrpc/$1/server/staging.replace"
@@ -193,6 +198,7 @@ main() {
         check_msg "$lang" "$want"
         want="$RANDOM"
         edit "$lang" "$want"
+        diff "$lang"
         update "$lang"
         sleep 25 # Mitigate against slow startup
         check_msg "$lang" "$want"
--- a/examples/hellogrpc/e2e-test.sh
+++ b/examples/hellogrpc/e2e-test.sh
@@ -59,6 +59,9 @@ function EXPECT_CONTAINS_PATTERN() {
 
 # Ensure there is an ip address for hell-grpc-staging:50051
 apply-lb() {
+    # We use `bazel build ... && bazel-bin/...` here in this file instead of
+    # `bazel run` directly in order to make sure that direct execution of the
+    # built output works as well
     logfail bazel build examples/hellogrpc:staging-service.apply
     logfail bazel-bin/examples/hellogrpc/staging-service.apply
 }
--- a/examples/hellogrpc/e2e-test.sh
+++ b/examples/hellogrpc/e2e-test.sh
@@ -126,7 +126,24 @@ edit() {
 
 diff() {
     logfail bazel build "examples/hellogrpc/$1/server:staging.diff"
-    logfail "bazel-bin/examples/hellogrpc/$1/server/staging.diff"
+    # We want diff to return 1 so we can't use logfail here
+    cmd="bazel-bin/examples/hellogrpc/$1/server/staging.diff"
+    echo "++ $cmd"
+    local out
+    local code
+    out=$("$cmd" 2>&1) && code=0 || code=$?
+    if [ $code -eq 1 ]; then
+        return 0
+    elif [ $code -eq 0 ]; then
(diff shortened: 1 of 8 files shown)


fejta:
/ok-to-test

Please add test coverage

https://github.com/bazelbuild/rules_k8s/blob/master/examples/hellohttp/e2e-test.sh
https://github.com/bazelbuild/rules_k8s/blob/master/examples/hellogrpc/e2e-test.sh

fejta:
Please rebase

Any chance you can get the e2e tests here passing? Thanks!

samschlegel:
Added to my tasks this week 😄

fejta:
Do things look any better this week? 😇

samschlegel:
heh. will do right now

examples/hellogrpc/e2e-test.sh
```diff
@@ -116,6 +116,11 @@ edit() {
     logfail "./examples/hellogrpc/$1/server/edit.sh" "$2"
 }
 
+diff() {
+    logfail bazel build "examples/hellogrpc/$1/server:staging.diff"
+    logfail "bazel-bin/examples/hellogrpc/$1/server/staging.diff"
```
samschlegel, line 121: is there a reason these don't use `bazel run`, which the hellohttp ones do? it doesn't seem like it's running in parallel so there shouldn't be lock contention. Is part of it just to test that direct execution works from outside the runfiles directory?

fejta: Right the goal is to ensure that both `bazel run //whatever` and `bazel build //whatever && bazel-bin/whatever` work correctly.

Since these are e2es it would be too slow to do both, so we just pick one for hellohttp and the other for hellogrpc

samschlegel: Gotcha. I'll add a comment in there in case someone else wonders why

samschlegel:
Rebased on master so tests will hopefully run 🤞

samschlegel:
Ah right. Diff returns exit codes based on if there was a diff so that needs to be handled 😅

samschlegel:
ah looks like some old flags are being passed into resolve, but previously the script was just dropping them.

fejta:
Hooray for test coverage :)

samschlegel:
I do wish this e2e test would work locally so I didn't have to publicly expose all this bad code I'm writing lol

fejta:
Don't worry, I don't tend to look at code until it passes tests :)

k8s/object.bzl
```diff
@@ -488,8 +508,12 @@ def k8s_object(name, **kwargs):
         **implicit_args
     )
 
-    _k8s_object(name = name, **kwargs)
-    _k8s_object(name = name + ".resolve", **kwargs)
+    resolve_args = dict(**kwargs)
+    if "args" in resolve_args:
+        resolve_args.pop("args")
```
samschlegel, line 513: I'm wondering if we should maybe just deprecate support for `args` and add a `kubectl_args` attribute so that it's clearer that this are only for the kubectl commands

k8s/go/cmd/resolver/resolver.go
```diff
@@ -37,6 +37,7 @@ var (
 	k8sTemplate       = flag.String("template", "", "The k8s YAML template file to resolve.")
 	substitutionsFile = flag.String("substitutions", "", "A file with a list of substitutions that were made in the YAML template. Any stamp values that appear are stamped by the resolver.")
 	allowUnusedImages = flag.Bool("allow_unused_images", false, "Allow images that don't appear in the JSON. This is useful when generating multiple SKUs of a k8s_object, only some of which use a particular image.")
+	noPush            = flag.Bool("no_push", false, "Don't push images after resolving digests.")
```
fejta, line 40: Will you avoid the `no` prefix?  Google code tends to consider `--nopush` equivalent to `--push-false` and is therefore unnecessarily confusing (even with the underscore).

```golang
flag.Bool("skip_push", false, "...")
flag.Bool("push", true, "...")
```

k8s/object.bzl
fejta: Makes sense but not totally sure it is worth it if others are already using args?

Can we document that `k8s_object` accepts `args` and/or `kubectl_args`?

k8s/go/cmd/resolver/resolver.go
fejta: Send a follow-up for this?

aignas:
I am really looking forward to using this.

@fejta, how often are you doing releases, when could I expect to have this in a tagged release?

PR #586, opened 2020-07-17
Resolve on delete

### LINKED ISSUE #182: delete does not call custom resolver / resolver at all
Thank you for your great rules.

I found problems with "*.delete" rule.
Unfortunately, resolver / customer resolver is not used on "delete".

I have following template:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: nginx-ingress
  namespace: <NAMESPACE>
...
```
And it can not be deleted by k8s rules.

So, would be great to call resolver on delete as well.

I know about k8s_defaults, but it is not suitable for me.

I have different k8s stands, like:
1. Personal Development stand (namespace name calculated based on AWS caller identity + git branch name)
2. Production stand (namespace fixed)

Right now I have the same rules for create k8s_object, and I am changing only resolver_args for my wrapper.

As result, first stand run with namespace like "oleg-tsarev-my_git_branch_name" and production with "production-project-name".

So, I forced to describe two different so different configuration, where the single difference - namespace name.

If you will fix resolver behaviour, I can use the same rules, and just change resolver.

mariusgrigoriu:
Fixes #182 

Invokes the resolver prior to reversing the order on delete.

To chain the resolver and reverser, the delete script is modified to pipe the output from the resolver to the reverser. The reverser is updated to read from standard in so it can be easily chained. As a result, the reverser has no need for args.

Since the reverser is called directly by the delete script, there's no need for a reverser script and its related rules. This should greatly simplify the flow. 

Finally, most args are the same, so they're extracted into common args to make it more clear as to what's different among these different rules.


fejta:
Thanks! Will you please add test coverage?

https://github.com/bazelbuild/rules_k8s/blob/master/examples/hellogrpc/e2e-test.sh
https://github.com/bazelbuild/rules_k8s/blob/master/examples/helloworld/e2e-test.sh

mariusgrigoriu:
Happy to add tests!

mariusgrigoriu:
Tests are passing. Ready for a review.

PR #626, opened 2020-12-04
fix gazelle

mariusgrigoriu:
Executing `bazel run //:gazelle` on master breaks the build due to changes introduced by `go_naming_convention` and due to the fact that the name of the `io_bazel_rules_docker` repository rule does not match what gazelle would generate based on the importpath. This PR should allow someone to run gazelle on this repository with breaking things.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -30,3 +30,6 @@ gazelle(
 
 # Make Gazelle ignore Go files in the examples directory used in e2e tests.
 # gazelle:exclude examples
+
+# rules_docker's BUILD files follow the go_default_library naming convention
+# gazelle:go_naming_convention_external go_default_library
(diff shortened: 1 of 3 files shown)


chases2 (approved):

mariusgrigoriu:
I think the test failure is unrelated. Every test across pull requests since Nov 18 has been timing out.

fejta:
Thanks, I believe the project where the e2e tests run have had the wrong restrictions put on it, causing these tests to fail. Need to find time to debug this.

mariusgrigoriu:
@fejta Any chance we could get the tests working again soon?

fejta:
How about if we just rename everything following gazelle's modern conventions?

PR #628, opened 2020-12-13
Resolver library

mariusgrigoriu:
There are times you may want a custom resolver, but not re-implement all of the behavior that exists in the one provided by rules_k8s. You could chain resolvers by executing the one from rules_k8s, but it would be nice if you could import it as a go package.

This PR moves the resolver into a library that custom Go resolvers can easily reuse to register flags and then pass execution to the rules_k8s resolver after performing custom processing.


chases2:
/ok-to-test
This looks like a fine refactoring to me, but I'd like another set of eyes on this.
/assign @fejta

fejta (approved):

PR #634, opened 2021-01-25
Allow rules_k8s targets to be invoked within other Bazel targets

EdSchouten:
The current logic for finding the runfiles directory only works when the
targets created by rules_k8s are invoked directly. Creating a
sh_binary() that chains multiple commands together doesn't work, as the
placement of the script relative to the runfiles directory will be
different.

This issue also affected rules_docker at some point in the past. This
change basically imports the new guess_runfiles() logic that
rules_docker uses.

### DIFF
--- a/k8s/apply.sh.tpl
+++ b/k8s/apply.sh.tpl
@@ -16,9 +16,14 @@
 set -euo pipefail
 
 function guess_runfiles() {
-    pushd ${BASH_SOURCE[0]}.runfiles > /dev/null 2>&1
-    pwd
-    popd > /dev/null 2>&1
+    if [ -d ${BASH_SOURCE[0]}.runfiles ]; then
+        # Runfiles are adjacent to the current script.
+        echo "$( cd ${BASH_SOURCE[0]}.runfiles && pwd )"
+    else
+        # The current script is within some other script's runfiles.
+        mydir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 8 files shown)


k8s/apply.sh.tpl
```diff
@@ -16,9 +16,14 @@
 set -euo pipefail
 
 function guess_runfiles() {
-    pushd ${BASH_SOURCE[0]}.runfiles > /dev/null 2>&1
-    pwd
-    popd > /dev/null 2>&1
+    if [ -d ${BASH_SOURCE[0]}.runfiles ]; then
+        # Runfiles are adjacent to the current script.
+        echo "$( cd ${BASH_SOURCE[0]}.runfiles && pwd )"
```
EdSchouten, line 21: This could also be written as:
```sh
(cd ${BASH_SOURCE[0]}.runfiles && pwd)
```

But I'd rather not deviate from what rules_docker does.

chases2 (approved):

EdSchouten:
Looks like CI failed due to flakiness, and I can't retrigger this myself. :-(

fejta:
> Looks like CI failed due to flakiness, and I can't retrigger this myself. :-(

You can most definitely trigger it yourself! See the `Rurun command` in the failure table and/or just type `/retest`

fejta:
Thank you!

EdSchouten:
Screenshot 2021-05-03 at 08 27 44

fejta:
The message says you need to wait for an `/ok-to-test` which you received.

PR #636, opened 2021-02-12
Handle kubectl diff exit code when diffs are found

goodspark:
kubectl diff exits with code 1 if a diff is found. Because of the exit flag, this causes the bazel kubectl diff action to exit early. So this attempts to handle that without losing general script-level exit code handling.


goodspark:
Hm. I'm not sure I the test failure is exactly related to this change:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/bazelbuild_rules_k8s/636/pull-rules-k8s-e2e/1363912632892919808#1:build-log.txt%3A33

```
++ FAIL: 1=bazel-bin/examples/hellogrpc/go/server/staging.create
$ /usr/bin/kubectl --kubeconfig= --cluster=gke_rules-k8s_us-central1-f_testing --context=gke_rules-k8s_us-central1-f_testing --user= --namespace=build-1363912632892919808 create -f -
error: unable to recognize "STDIN": no matches for kind "Deployment" in version "apps/v1beta1"
```

goodspark:
Yeah given the error is `no matches for kind "Deployment" in version "apps/v1beta1"`, this makes me think the k8s cluster that's being used for testing was upgraded without appropriate updates to the test files in this repo. `apps/v1beta1` was removed in k8s 1.18. Perhaps the testing k8s cluster was upgraded to this or above?

chases2 (approved):

fejta:
Please fix tests

fejta:
This seems to be failing tests, would you like to try and fix the tests or close out this PR?

PR #638, opened 2021-02-25
Try to fix broken tests

### LINKED ISSUE #637: Tests appear to be using outdated apiVersions
From https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/bazelbuild_rules_k8s/636/pull-rules-k8s-e2e/1363959740941471744#1:build-log.txt%3A31

> ++ FAIL: 1=bazel-bin/examples/hellogrpc/go/server/staging.create
$ /usr/bin/kubectl --kubeconfig= --cluster=gke_rules-k8s_us-central1-f_testing --context=gke_rules-k8s_us-central1-f_testing --user= --namespace=build-1363959740941471744 create -f -
error: unable to recognize "STDIN": no matches for kind "Deployment" in version "apps/v1beta1"

And based on the version info:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/bazelbuild_rules_k8s/636/pull-rules-k8s-e2e/1363959740941471744#1:build-log.txt%3A65

I think the test tooling inadvertently got upgraded to using kubectl 1.18+ ([where the `apps/v1beta1` apiVersion was removed](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#kube-apiserver-1)) without updates to the tests.

I think the fix should be just changing the various k8s libsonnet files to use up-to-date apiVersions.

goodspark:
Fixes #637 (maybe?)

### DIFF
--- a/examples/deployment.libsonnet
+++ b/examples/deployment.libsonnet
@@ -14,7 +14,7 @@
 
 {
    Simple:: function(name, containerSpec) {
-      "apiVersion": "apps/v1beta1",
+      "apiVersion": "apps/v1",
       "kind": "Deployment",
       "metadata": {
          "name": name
--- a/examples/deployment.libsonnet
+++ b/examples/deployment.libsonnet
@@ -21,6 +21,11 @@
       },
       "spec": {
          "replicas": 1,
+         "selector": {
+            "matchLabels": {
+               "app": name,
+            },
+         },
          "template": {
             "metadata": {
                "labels": {
--- a/examples/hellogrpc/deployment.yaml
+++ b/examples/hellogrpc/deployment.yaml
@@ -4,6 +4,9 @@ metadata:
   name: hello-grpc-staging
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      app: hello-grpc-staging
   template:
     metadata:
       labels:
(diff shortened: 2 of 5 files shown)


goodspark:
@googlebot I fixed it.

goodspark:
Looks like the tests got further than before so I think the linked issue is fixed, but I think there's something else in the tests that's broken. It's been hanging on the same line for a while. Looking back for older tests, it seems there's some kind of timeout issue (tests taking 2hrs+) even going as far back as December: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/bazelbuild_rules_k8s/619/pull-rules-k8s-e2e/1334302328328556544

I think some next steps would be:
- Merge this if owners think it looks good
- Open another issue tracking whatever it making the tests hang and timeout

fejta:
The test is failing in part because it is no longer allowed to expose the services it is creating, and we've not yet had the time to  refactor these tests to use `kubectl port-forward` or some other mitigation.

UebelAndre:
@fejta do you have a timeline on when this will be fixed? It currently blocks all PRs to this project, correct?

fejta:
Correct. No timeline on fix.

PR #639, opened 2021-03-02
Add support for Bazel 3.0.0

UebelAndre:
This PR adds Bazel 3.0.0 support by updating various dependencies.


UebelAndre:
@chases2 Any way to help resolve 
```
ERROR: The project you're trying to build requires Bazel 4.0.0 (specified in /home/prow/go/src/github.com/bazelbuild/rules_k8s/.bazelversion), but it wasn't found in /usr/local/lib/bazel/bin.
```
?

fejta:
https://github.com/bazelbuild/rules_k8s/blob/002b365cfd200192569fa1ba2b6b3178b693989f/images/gcloud-bazel/Makefile#L18-L19

The image has 2.2.0 and 3.0.0. Can we start by upgrading to 3.0.0? Changing to `NEW = 4.0.0` fails:

```
Step #0: Step 4/14 : FROM launcher.gcr.io/google/bazel:${NEW_VERSION}
Step #0: manifest for launcher.gcr.io/google/bazel:4.0.0 not found: manifest unknown: Failed to fetch "4.0.0" from request "/v2/google/bazel/manifests/4.0.0".
Finished Step #0
ERROR
ERROR: build step 0 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

ERROR: (gcloud.builds.submit) build e75d50d1-1986-4e4d-adb3-1ea8269c42f2 completed with status "FAILURE"
```

UebelAndre:
> The image has 2.2.0 and 3.0.0. Can we start by upgrading to 3.0.0? Changing to `NEW = 4.0.0` fails:

Trying 3.0.0 but it feels kinda bad given it's already a year old. In this PR I added a test to run on `2.2.0` to ensure previous versions of Bazel still work. The thing we don't know is if newer ones also work (but we know whatever runs in BuildKite does though). I think it'd be beneficial to setup `min supported` and `latest` e2e tests.

fejta (approved):

fejta:
If you give me a valid tag I can use for `launcher.gcr.io/google/bazel:4.0.0` then I can create an image that has bazel 4.x.y and 3.0.0 (at which point we should probably update the min supported version to 3.0.0)

UebelAndre:
> If you give me a valid tag I can use for `launcher.gcr.io/google/bazel:4.0.0` then I can create an image that has bazel 4.x.y and 3.0.0 (at which point we should probably update the min supported version to 3.0.0)

I actually have no idea who would make that or how to get one. Is there a Googler you could reach out to?

UebelAndre:
Oh, I messed up the commit message, 3.0.0 is not an LTS release :(

fejta:
I believe the latest that exists there is 3.5.0

UebelAndre:
> I believe the latest that exists there is 3.5.0

Right but who publishes those? I feel like a 4.0.0 image is warranted but I feel that's less urgent since 4.0.0 was confirmed to work in this PR.

Issue #642, opened 2021-03-12
e2e tests are broken


fejta:
An example failure: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/bazelbuild_rules_k8s/638/pull-rules-k8s-e2e/1366496761295671296

Problem is that no longer has the ability to create the externally-accessible load balancers that the e2e tests currently depend on, making this part of the e2e tests fail.

Tests create an external LB: 
https://github.com/bazelbuild/rules_k8s/blob/f1c6399cdd691b7aca90073398e8f690ec8992c6/examples/hellohttp/e2e-test.sh#L45-L48

When then wait for an IP address:https://github.com/bazelbuild/rules_k8s/blob/f1c6399cdd691b7aca90073398e8f690ec8992c6/examples/hellohttp/e2e-test.sh#L51-L61

But we will never get one:
https://github.com/bazelbuild/rules_k8s/blob/f1c6399cdd691b7aca90073398e8f690ec8992c6/examples/util.sh#L60-L70

Basically we need to either fix the project to regain this ability or else update the tests to no longer require it.

For the latter situation, I would expect to be able to use `kubectl port-forward` to provide a localhost IP address that connects to the service on the cluster, probably with minor changes to the [https://github.com/bazelbuild/rules_k8s/blob/f1c6399cdd691b7aca90073398e8f690ec8992c6/examples/hellohttp/service.yaml](service).

f0rmiga:
I'd starting by asking _why are we poking the services at all_?
If we trust the k8s API (on a correctly configured cluster), I don't see a need to exercise the examples business logic.
We can make assertions based on the k8s events.

If I'm missing something and _we **must** exercise the business logic_, then `kubectl port-forward` indeed sounds a good candidate.

chrislovecnm:
@f0rmiga can we change the CI to not create services? I would agree that we do not need to do this because we are essentially testing the k8s API and not rules_k8s.  We need to validate that we are communicating the API correctly and that we are installing valid k8s objects, but not need to create services.

f0rmiga:
@chrislovecnm I don't think _creating_ the service is the problem. We can change the assertion from _poking the service_ to _asserting all the resources were created as expected_.

Working on it now.

chrislovecnm:
> Problem is that no longer has the ability to create the externally-accessible load balancers that the e2e tests currently depend on, making this part of the e2e tests fail.

[...]

PR #652, opened 2021-04-21
WIP: Reimplement hellogrpc test

chases2:
xref #642

### DIFF
--- a/examples/util.sh
+++ b/examples/util.sh
@@ -65,6 +65,7 @@ get_lb_ip() {
     ip_var='{.spec.clusterIP}'
   fi
 
-  kubectl --namespace="${namespace}" get service $2 \
-    -o jsonpath=$ip_var
+  kubectl --namespace="${namespace}" port-forward $2 :50000 &
+
+  echo "50000"
 }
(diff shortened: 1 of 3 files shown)


examples/util.sh
```diff
@@ -65,6 +65,7 @@ get_lb_ip() {
     ip_var='{.spec.clusterIP}'
   fi
 
-  kubectl --namespace="${namespace}" get service $2 \
-    -o jsonpath=$ip_var
+  kubectl --namespace="${namespace}" port-forward $2 :50000 &
```
fejta, line 68: `kubectl help port-forward` suggests `:50000` chooses a random local port and forwards it to port `50000` on `$2`

Reading these services does not suggest they are listening on port 50000: 

https://github.com/bazelbuild/rules_k8s/blob/002b365cfd200192569fa1ba2b6b3178b693989f/examples/hellohttp/service.yaml#L11-L13

Also I'm not entirely sure but I suspect you may need to change the service type:
https://github.com/bazelbuild/rules_k8s/blob/002b365cfd200192569fa1ba2b6b3178b693989f/examples/hellohttp/service.yaml#L16

fejta: Looks like hellogrpc listens on a different port:

https://github.com/bazelbuild/rules_k8s/blob/002b365cfd200192569fa1ba2b6b3178b693989f/examples/hellogrpc/service.yaml#L9-L11

chases2:
Abandoning in favor of https://github.com/bazelbuild/rules_k8s/pull/655

PR #656, opened 2021-04-30
Update rules_docker

fejta:
Also vert 73f62de by reverting 934d2a1f7859987967901fab68c21855ff923b91

/cc @UebelAndre

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -49,9 +49,5 @@ platforms:
     build_targets:
     - "--"
     - "..."
-    - "-//images/gcloud-bazel:gcloud_push"
-    - "-//images/gcloud-bazel:gcloud_installer"
-    - "-//images/gcloud-bazel:gcloud_install"
-    - "-//images/gcloud-bazel:gcloud-layer"
     test_targets:
     - "..."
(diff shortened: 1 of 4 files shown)


fejta:
(@UebelAndre any chance you can either respond to the `@googlebot I consent` message or else patch in the revert commit into your PR)

.bazelci/presubmit.yml
```diff
@@ -44,15 +38,20 @@ platforms:
...
-    # TODO(nlopezgi): remove once issue above is fixed
-    - "--action_env=PATH=/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/tmp/bin/"
+  min_supported_version:
+    name: "Minimum Supported Version"
+    bazel: "2.2.0"
+    platform: ubuntu1604
+    build_targets:
+    - "--"
+    - "..."
+    - "-//images/gcloud-bazel:gcloud_push"
```
UebelAndre, line 52: I forget, why were these disabled? Seems like they should be enabled

UebelAndre:
@googlebot I consent

.bazelci/presubmit.yml
fejta: Probably didn't build at the time or something, let's see if they work now.

fejta:
> @googlebot I consent

Unfortunately I think it might need to be exactly `@googlebot I consent.` (with period, which I missed initially)

UebelAndre:
It says CLA is good to go but

@googlebot I consent.

fejta:
oh sorry my bad!

fejta:
(I'm used to other repos where there's a label and not just the context, thanks!)

UebelAndre:
> (I'm used to other repos where there's a label and not just the context, thanks!)

Thanks for putting this PR together and fixing CI 😄

chases2 (approved):

PR #669, opened 2021-10-12
Update rules_docker to v0.20.0

mishas:
New version of rules_docker was release, and it has a change that breaks rules_k8s.


mishas:
Hello @chizhg and @fejta ,

I'm not really sure about how the "praw" process works, but from the link in "The pull request process is described here" mentioned above it says that the PR will get automatically merged if all tests pass and the PR has the "LGTM" and "approved" labels.

I'm not sure if it's a bug in the automated system, or is there anything else that needs to be done here?

Thank you.

chizhg:
/assign @fejta 
Shall we configure Tide for this repo?

fejta:
We should first move it to oss-prow.knative.dev

PR #675, opened 2021-11-08
Do not use constraints from bazel_tools

### LINKED ISSUE #674: Do not use constraints from bazel_tools
Constraints from bazel_tools should not be used, instead the ones from https://github.com/bazelbuild/platforms should be used. In Bazel 5.0 this will be flipped and already now to prepare the migration to Bazel 5.0 someone depending on rules_docker cannot flip the flag because rules_docker still uses bazel_tools.
More info in: bazelbuild/bazel#8622

limdor:
Fixes #674

### DIFF
--- a/toolchains/kubectl/BUILD
+++ b/toolchains/kubectl/BUILD
@@ -35,8 +35,8 @@ kubectl_toolchain(
 toolchain(
     name = "kubectl_linux_amd64_toolchain",
     target_compatible_with = [
-        "@bazel_tools//platforms:linux",
-        "@bazel_tools//platforms:x86_64",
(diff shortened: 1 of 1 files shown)


limdor:
@fejta any updates on this? Without this it is not possible to use Bazel 5 with the default flags.

limdor:
@fejta @chases2 gentle ping

meteorcloudy (approved):

meteorcloudy:
@chases2 Can you or any other owner of rules_k8s take a look before merging this?

chases2 (approved):

limdor:
Thanks a lot!

PR #678, opened 2022-01-19
Proposal: add arg to ignore image digest resoultion

fredr:
When using git hashes as image tags we are fairly protected against images being mutated (especially where we build our own images and are in control of how they are being published).
It can then be nice to have the git hash as tag to be able to quickly see what has been deployed.

This adds an option for the resolver to ignore certain images from digest resolution, based on their prefix.

example usage
```
k8s_object(
    name = "k8s",
    resolver_args = [
        "-ignore_image_prefix=gcr.io/google-containers/pause",
        "-ignore_image_prefix=gcr.io/a-whole-repositoy/",
    ],
    kind = "...",
    template = "deployment.yaml",
)
```

~PS: isn't it `resolver_args` that should be forwarded to the resolver, not `args`? or is the documentation outdated?~ (this was on k8s_objects, not on k8s_object, my bad)

### DIFF
--- a/k8s/go/pkg/resolver/resolver.go
+++ b/k8s/go/pkg/resolver/resolver.go
@@ -21,24 +21,26 @@ import (
 
 // Flags defines the flags that rules_k8s may pass to the resolver
 type Flags struct {
-	ImgChroot         string
-	K8sTemplate       string
-	SubstitutionsFile string
-	AllowUnusedImages bool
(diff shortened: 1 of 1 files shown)


fredr:
I ended up writing a custom resolver that doesn't resolve to digest. If this is not something people generally are looking for we can just close this.

wdyt @chases2?

Issue #681, opened 2022-02-01
Bazel builds started failing due to 


SGudbrandsson:
Bazel builds are now failing due to a failed checksum error.
This started happening this morning.

https://github.com/bazelbuild/rules_k8s/blob/d05cbea5c56738ef02c667c10951294928a1d64a/k8s/k8s.bzl#L54

Anyone else experiencing this?

```bazel
WARNING: Download from https://github.com/yaml/pyyaml/archive/5.1.2.zip failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException Checksum was ba59d7e97eb131d8f8f52d19cb124bb67772f4c7f4d14cb2919deb885ef8c572 but wanted e9df8412ddabc9c21b4437ee138875b95ebb32c25f07f962439e16005152e00e
ERROR: An error occurred during the fetch of repository 'com_github_yaml_pyyaml':
   java.io.IOException: Error downloading [https://github.com/yaml/pyyaml/archive/5.1.2.zip] to /home/sigurdur/.cache/bazel/_bazel_sigurdur/71ecaec01697b935e8cbde2c3c9da6e5/external/com_github_yaml_pyyaml/5.1.2.zip: Checksum was ba59d7e97eb131d8f8f52d19cb124bb67772f4c7f4d14cb2919deb885ef8c572 but wanted e9df8412ddabc9c21b4437ee138875b95ebb32c25f07f962439e16005152e00e
ERROR: Evaluation of query "filter('\.apply', attr('tags', 'upload-assets', attr('tags', 'dev', allrdeps(set()))))" failed due to BUILD file errors
```

rajatvig:
https://github.com/bazelbuild/rules_k8s/pull/683 should fix it.

jarimayenburg:
Yup, same thing here

jarimayenburg:
Might be onto something here.

bazelbuild/rules_k8s uses the URL https://github.com/yaml/pyyaml/archive/5.1.2.zip, expecting checksum e9df8412ddabc9c21b4437ee138875b95ebb32c25f07f962439e16005152e00e while the file has checksum ba59d7e97eb131d8f8f52d19cb124bb67772f4c7f4d14cb2919deb885ef8c572

However, using URL https://github.com/yaml/pyyaml/archive/refs/tags/5.1.2.zip still yields checksum e9df8412ddabc9c21b4437ee138875b95ebb32c25f07f962439e16005152e00e.

[...]

PR #683, opened 2022-02-01
Support M1

rajatvig:


k8s/k8s.bzl
```diff
@@ -51,7 +51,7 @@ def k8s_repositories():
         http_archive,
         name = "com_github_yaml_pyyaml",
         build_file_content = _com_github_yaml_pyyaml_build_file,
-        sha256 = "e9df8412ddabc9c21b4437ee138875b95ebb32c25f07f962439e16005152e00e",
+        sha256 = "ba59d7e97eb131d8f8f52d19cb124bb67772f4c7f4d14cb2919deb885ef8c572",
```
fejta, line 54: Any insight into why/when the sha changed?

rajatvig: Seems like it has been fixed. Will revert. [This](https://twitter.com/tgummerer/status/1488493440103030787?s=20&t=DaupVstH5BiSJd0BgLSa_Q) has the info. There is chat in the [issue](https://github.com/bazelbuild/rules_k8s/issues/681) as well

fejta:
I suspect we'll need to wait for various caches to become consistent again.

rajatvig:
> I suspect we'll need to wait for various caches to become consistent again.

Very likely but unsure. To test, just downloaded the archive and it has the new SHA. Ran a clean bazel build with a local patch for it and that too used the new SHA.

moriarty:
> > I suspect we'll need to wait for various caches to become consistent again.
> 
> Very likely but unsure. To test, just downloaded the archive and it has the new SHA. Ran a clean bazel build with a local patch for it and that too used the new SHA.

```bash
while true; do curl --silent -L https://github.com/yaml/pyyaml/archive/5.1.2.zip | shasum -a 256; sleep 2; done
```

Try that and see if you get consistent results. It seems to vary based on location and ISP... hopefully it trickles through all the various CDNs

rajatvig:
@fejta From what it looks like, most caches have now got it fixed.

fejta:
Can you drop the last two commits from this PR?

rajatvig:
@fejta Dropped the two commits but I do not have rights to rebuild

rajatvig:
@fejta Could you retrigger the tests?

fejta:
Mostly superseded by https://github.com/bazelbuild/rules_k8s/pull/697

Issue #696, opened 2022-05-13
kubectl toolchain for Apple M1 (cpu: arm64) on os:macos


lomeshpatel:
There doesn't appear to be support for `kubectl` toolchain for Apple M1 processor. Here's what I'm running.

```
$ uname -a
Darwin local 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000 arm64
```

When running `bazel run` commands I get the following error:
```
$ bazel run :echo.create --toolchain_resolution_debug=@io_bazel_rules_k8s//toolchains/kubectl:toolchain_type
INFO: Build option --toolchain_resolution_debug has changed, discarding analysis cache.
INFO: ToolchainResolution:     Type @io_bazel_rules_k8s//toolchains/kubectl:toolchain_type: target platform @local_config_platform//:host: Rejected toolchain @k8s_config//:toolchain; mismatching values: linux, x86_64
INFO: ToolchainResolution:     Type @io_bazel_rules_k8s//toolchains/kubectl:toolchain_type: target platform @local_config_platform//:host: Rejected toolchain @k8s_config//:toolchain; mismatching values: x86_64
INFO: ToolchainResolution:     Type @io_bazel_rules_k8s//toolchains/kubectl:toolchain_type: target platform @local_config_platform//:host: Rejected toolchain @k8s_config//:toolchain; mismatching values: windows, x86_64
INFO: ToolchainResolution:   Type @io_bazel_rules_k8s//toolchains/kubectl:toolchain_type: target platform @local_config_platform//:host: No toolchains found.
ERROR: /Users/lomesh/myprojects/bazel/examples/k8s/BUILD:3:17: While resolving toolchains for target //:echo.create: no matching toolchains found for types @io_bazel_rules_k8s//toolchains/kubectl:toolchain_type
ERROR: Analysis of target '//:echo.create' failed; build aborted:
INFO: Elapsed time: 0.182s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 250 targets configured)
```

I've made the necessary changes and will be submitting a PR to fix this.

I've also made `@io_bazel_rules_k8s//toolchains/kubectl:toolchain_type` `public` to allow for extensions, in case the similar situations arise for other OS and/or CPU.

PS: I'm fairly new to Bazel, so please point me in the right direction if need be.

[...]

PR #697, opened 2022-05-13
Added toolchain support for Apple M1 and made toolchain_type public to allow extensions

lomeshpatel:
1. Added and registered `kubectl_macos_arm64_toolchain`
2. Changed the name of `kubectl_osx_toolchain` to `kubectl_macos_x86_64_toolchain` and added an alias for backwards compatibility. Also updated the registration.
3. Made `toolchain/kubectl` package `public` to allow for extensions for this toolchain_type

#696


fejta (approved):

lomeshpatel:
@fejta I noticed that one of the checks failed, is that expected or do I need to fix something?

Also, can you please point me to the release process? I just would like to have some understanding of when these changes will be available.

Issue #723, opened 2023-04-05
Drop dependency on rules_docker


alexeagle:
Because it is in "minimal maintenance mode"
https://github.com/bazelbuild/rules_docker#status

I think the usages of rules_docker in this repo are minimal and could be easily satisfied without the dependency.

Related to #722

mrmeku:
This is the only tricky bit of dropping rules_docker entirely
https://github.com/bazelbuild/rules_k8s/blob/7fe8176fecc0082753c97deaf0f18d434d5a8b1a/k8s/object.bzl#L68

Other than that, rules_docker can be easily replaced by `rules_oci` for packaging or by inlining helper functions which are currently imported from `rules_docker`. Which @alexeagle has already mostly taken care of with a PR.

Translating the above into something that's compatible with `rules_oci` will be a bit different since `oci_image` produces a directory artifact as its only output where as rules_docker seems to have shipped with a provider which made it easier to consume in the bazel layer. 

However, that artifact will always follow the OCI image-spec 
https://github.com/opencontainers/image-spec/blob/main/image-layout.md

It seems practical to teach rules_k8s how to consume directory artifacts that fit that layout somehow. Certainly, the manifest file and the blobs directory of the directory artifact need to be translated to the bazel layer at some point in this process. Perhaps rules_oci might consider shipping with a helper method to make this translation easier for downstream consumers @alexeagle ?

realtimetodie:
I would take care of this, but I think that the PR would never be reviewed. Unfortunately, this repository appears to be unmaintained.

PR #725, opened 2023-04-07
Inline utils from rules_docker go pkg

mrmeku:
For #723

### DIFF
--- a/k8s/go/pkg/resolver/resolver.go
+++ b/k8s/go/pkg/resolver/resolver.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 
 	"github.com/bazelbuild/rules_docker/container/go/pkg/compat"
-	"github.com/bazelbuild/rules_docker/container/go/pkg/utils"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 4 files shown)


k8s/go/pkg/resolver/resolver.go
```diff
@@ -12,7 +12,6 @@ import (
 	"strings"
 
 	"github.com/bazelbuild/rules_docker/container/go/pkg/compat"
```
alexeagle, line 14: what about this one, should it be inlined too? I guess you'll have a later PR to drop the go_repository that fetches rules_docker code. I don't know what is the maintenance story for rules_k8s so maybe we should wait-and-see if we get any responses.

alexeagle (commented):
awesome nice to see you again @mrmeku :)

k8s/go/pkg/resolver/resolver.go
mrmeku: Yeah, maybe that one should also be inlined as part of this change. I had held off since that particular import needs to be translated to understand the OCI image format rather than the format rules_docker is outputting today. But that translation could easily be done in a followup PR after inlining in this one

alexeagle:
@chases2 hi, are you the maintainer here?

